Issue #7, opened 2016-05-04
Mismatch between port arguments to the asyncio and uvloop loop.create_server method 


claws:
There appears to be a discrepancy in valid `port` arguments to the event loop's `create_server` method. With `asyncio` I can use `0` and `None` to obtain a server bound to an ephemeral port. With `uvloop` it appears that `None` is invalid and raises an error from the `_getaddrinfo` method.

```
File "uvloop/loop.pyx", line 1026, in create_server (uvloop/loop.c:19287)
File "uvloop/loop.pyx", line 505, in uvloop.loop.Loop._getaddrinfo (uvloop/loop.c:11976)
TypeError: port must be a str, bytes or int
```
I came across this because in my use case I create many servers on ephemeral ports that then register their actual bound endpoint details with a service discovery mechanism. In my `asyncio` based implementation I leave the port at its default value (e.g. `port=None`) but this failed when I was testing switching over to `uvloop`. I was able to modify my code to explicitly use `port=0` to get it to work with both event loops.

It would be nice if `uvloop` was a true drop in replacement for `asyncio`. I expect that a check in the [_getaddrinfo](https://github.com/MagicStack/uvloop/blob/master/uvloop/loop.pyx#L495) method could accomodate this discrepancy.

I have created a short [gist](https://gist.github.com/claws/213d0f86a6ef9f6f92b82c5fcb97ae95) that can be used to demonstrate the difference.

1st1:
The PR is merged, closing this one. Thanks a lot for the contribution!

PR #8, opened 2016-05-04
fix port value mismatch in create_server

claws:
This pull request provides a fix for the issue outlined in #7.

The `asyncio` `loop.create_server` method accepts `port=None` as a valid argument but `uvloop` raises an exception. To approach true drop in replacement parity `uvloop` should ideally allow this argument and create an ephemeral port the same as `asyncio` does.

This change also provides a unit test to verify the functionality.


1st1:
LGTM. Will merge later today.

1st1:
Merged by hand. I've also published uvloop v0.4.11 with your fix in. Thanks a lot!

PR #18, opened 2016-05-08
test code cleanups

claws:
The pull request consolidates all test code into the ``tests`` directory (e.g. ``_testbase.py``) and removes the assumption that tests are only ever run from the top level directory. This change allows the tests to be run with either ``unittest`` or ``pytest`` from either the top level directory or from within the tests directory. 

I typically use the standard library ``unittest`` package to run unit tests and I was using it with uvloop as they all seemed like typical ``unittest`` style tests to me and I saw no mention of using ``pytest``. I tend to use ``unittest`` because it is in the standard library which just reduces the need for one extra dependency. 

Using the common ``unittest`` approach of running individual unit tests from within the tests directory I was getting test failures. The same tests would pass when run using the ``make test`` rule. This discrepancy was because some paths (e.g. to certs) were hard coded to always assume tests are run from the top level directory. 

Via another PR I see that you use ``pytest``, though I don't think there is anything ``pytest`` specific in the tests. This change allows either approach to be used.

### DIFF
--- a/tests/_testbase.py
+++ b/tests/_testbase.py
@@ -62,8 +62,8 @@ class BaseTestCase(unittest.TestCase):
 
 def _cert_fullname(name):
     fullname = os.path.join(
-        os.path.dirname(os.path.dirname(__file__)),
-        'tests', 'certs', name)
+        os.path.dirname(__file__),
+        'certs', name)
     assert os.path.isfile(fullname)
     return fullname
 
--- a/tests/test_aiohttp.py
+++ b/tests/test_aiohttp.py
@@ -8,7 +8,7 @@ else:
 
 import unittest
 
-from uvloop import _testbase as tb
+import _testbase as tb
 
 
 class _TestAioHTTP:
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -7,7 +7,7 @@ import uvloop
 import unittest
 
 from unittest import mock
-from uvloop._testbase import UVTestCase, AIOTestCase
+from _testbase import UVTestCase, AIOTestCase
 
 
 class _TestBase:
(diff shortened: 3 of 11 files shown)


1st1:
I've just discovered the `pip -e` thing. Essentially, in your uvloop development venv, you do the following:

```shell
$ cd uvloop
$ pip -e .
```

After that you can run `python -m unittest` from the `uvloop/tests` directory. py.test can be used without a `$ PYTHONPATH=. py.test` hack, and can also be used from any dir.

I think we just need to document the `pip -e` as a recommendation for uvloop venv setup.

claws:
Your suggestion certainly looks like it would help in the use case that caused me to raise this PR. A minor clarification; I found that I had to use ``pip install -e .``.

It seems a little uncommon (admittedly there is variability between projects) to have pure test utils in the main package. My thought space on this topic is reducing the barrier to entry for any potential future contributors. What was the reason for putting some test utilities in the main uvloop package? 

It might be worth putting a note somewhere (perhaps in ``_testbase.py``) about your design decision regarding this aspect. I can foresee future contributors wondering why test utils are in the main package and making the same suggestion I have.

1st1:
> It seems a little uncommon (admittedly there is variability between projects) to have pure test utils in the main package. My thought space on this topic is reducing the barrier to entry for any potential future contributors. What was the reason for putting some test utilities in the main uvloop package?

TBH I didn't put a lot of thought into that.  Having some test harness in a module inside the main package is something we did in asyncio, see `asyncio/test_utils.py` for instance.

In this particular PR, I don't like the `import _testbase` line.  I know why it works, but still it looks confusing to me.

claws:
Yeah I was thinking of ``asyncio`` when I mentioned test utils variability.

OK, I'll close this out then.

PR #19, opened 2016-05-08
add docs

claws:
This pull request adds documentation to the uvloop project. A live demonstration can be found [here](http://uvloop-demo.readthedocs.io/en/docs/). The documentation uses Sphinx and can be linked to [readthedocs](https://readthedocs.org/) where many Python packages host their documentation.

The documentation contains a main page containing a brief overview. Then there is a user guide (install and using uvloop), a developers guide (getting the source, building, testing, docs) and the API page. The makefile has been updated to make it easy to build the docs.

I have also updated the loop.pyx file with comments similar to their asyncio counterparts. This file also contains a compiler directive (embedsignature) that allows Sphinx to pull the docs trings from Cython modules.

This work builds upon an earlier PR (which I closed because I stuffed up the git repo with a rebase that went haywire) in which there were some existing review comments. The content of this PR has addressed some of these comments, a summary is presented below:

- Added requirements.dev.txt to contain Cython.
  - if you think the project should use pytest for testing should the file also contain pytest?
- Added a link to the ReadTheDocs docs in the README.rst.
- Added a link to the magic.io uvloop blog post in the docs index page.
- Added a reference for using pytest as well as unittest.
- Added a sentence to the dev docs stating that any deviation from the reference `asyncio` behavior is considered a bug.
- Removed reference to needing Cython for user installs.
- Added ``install Cython`` note to dev docs.
- Sphinx docs/projects typically use a logo. This project does not have a logo yet.
  - one consequence is that the docs theme will use the title `uvloop` if no image is provided. It looks a bit clumsy but if you add a logo image then the title can be muted so it should be fine.
- embedsignature is used to allow sphinx to pull docs strings from the Cython modules.


docs/index.rst
```diff
@@ -0,0 +1,66 @@
+
+.. raw:: html
```
1st1: Let's just copy the shields code from README.rst. No need to center them.

docs/user/index.rst
```diff
@@ -0,0 +1,40 @@
+User Guide
+==========
+
+This section of the documentation provides information about how to install
+uvloop as well as how to use uvloop.
```
1st1: I'd rephrase to:

    This section of the documentation provides information about 
    how to use uvloop.

docs/conf.py
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/env python3
```
1st1: Can we remove all the commented code etc?

1st1:
Another thought. I'd like this to be split in two PRs: one that adds docstrings to the Loop methods; and one that sets up docs/sphinx/etc.

docs/dev/index.rst
```diff
@@ -0,0 +1,175 @@
...
+    The best way to work on `uvloop` is to create a virtual env, so that
+    you'll have Cython and Python commands pointing to the correct
+    tools.
+
+    .. code-block:: console
+
+        $ python3 -m venv myvenv
+        $ cd myvenv
+        $ source bin/activate
+        $ cd ..
```
1st1: Let's add a note about `pip -e .`

docs/dev/index.rst
```diff
@@ -0,0 +1,175 @@
...
+
+.. code-block:: console
+
+    $ cd uvloop
+    $ make test
+
+Alternatively, you can run the tests from the ``tests`` directory using the
+standard library `unittest` approach. When using this approach you must
+specify the ``PYTHONPATH`` so that the `uvloop` package can be resolved from
+down in the ``tests`` directory.
```
1st1: Again, `pip -e` solves the `PYTHONPATH` problem.

docs/dev/index.rst
```diff
@@ -0,0 +1,175 @@
...
+
+.. code-block:: console
+
+    $ cd docs/
+    $ python3 -m http.server
+
+Then open http://localhost:8000 in a browser and navigate to
+``_build/html/index.html``.
+
+.. note::
```
1st1: I don't think we need to go into too much detail here... Let's keep the docs as concise as possible ;)

requirements.dev.txt
```diff
@@ -0,0 +1,2 @@
+Cython
```
1st1: Let's specify the min required versions of Cython and sphinx here.

1st1:
Alright, I've left some feedback. Generally looks good, aside from a few nits. Let's break this PR into two though.

docs/Makefile
```diff
@@ -0,0 +1,192 @@
+# Makefile for Sphinx documentation
```
1st1: Maybe we don't need this Makefile?  Would it be possible to just call sphinx directly (using `python -m sphinx`) from the `uvloop/Makefile`?

claws:
I think that this addresses all the review comments.

- removed the loop.pyx comments update to a separate PR.
- loop.pyx now contains just the Sphinx related change supporting docs.
- same shields as README
- updated user doc with suggested re-write.
- trimmed Sphinx conf.py to minimal content (removed commented out stuff)
- added reference to using ``pip install -e .`` instead of using PYTHONPATH.
- trimmed testing section
- trimmed the docs section as suggested.
- updated requirements.txt with package versions
- Removed Sphinx Makefile with a call in from the main Makefile directly to
  the sphinx-build call.

docs/dev/index.rst
claws: Added it down in the tests section.

docs/conf.py
claws: Done

docs/index.rst
claws: Back to two shields.

1st1:
Looks good! Will merge it today.

1st1:
Committed in 8511fec1f9a610b0b26547a1620ce034ad9d9947. Thanks a lot!

PR #22, opened 2016-05-09
Allow dynamic linking to system libuv

mtorromeo:
Added build_ext option --use-system-libuv to switch between dynamic and static linking.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,17 @@ LIBUV_DIR = os.path.join(os.path.dirname(__file__), 'vendor', 'libuv')
 
 
 class libuv_build_ext(build_ext):
+    build_ext.user_options.extend([
+        ("use-system-libuv", None,
+         "Use the system provided libuv, instead of the bundled one")
+    ])
+
+    build_ext.boolean_options.extend(["use-system-libuv"])
+
+    def initialize_options(self):
+        build_ext.initialize_options(self)
+        self.use_system_libuv = 0
+
     def build_libuv(self):
         env = os.environ.copy()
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -17,6 +17,16 @@
 
 
 class libuv_build_ext(build_ext):
+    build_ext.user_options.extend([
+        ("use-system-libuv", None, "Use the system provided libuv, instead of the bundled one")
```
1st1: This line is too long.

mtorromeo: Fixed

setup.py
```diff
@@ -35,14 +45,17 @@ def build_libuv(self):
...
-        libuv_lib = os.path.join(LIBUV_DIR, '.libs', 'libuv.a')
-        if not os.path.exists(libuv_lib):
-            self.build_libuv()
-        if not os.path.exists(libuv_lib):
-            raise RuntimeError('failed to build libuv')
-
-        self.extensions[-1].extra_objects.extend([libuv_lib])
-        self.compiler.add_include_dir(os.path.join(LIBUV_DIR, 'include'))
+        if self.use_system_libuv:
+            self.compiler.add_library('uv')
```
1st1: Can you also add this snippet of code here, to support Mac OS X + macports:

```python
            if sys.platform == 'darwin' and \
                    os.path.exists('/opt/local/include'):
                # Support macports on Mac OS X.
                self.compiler.add_include_dir('/opt/local/include')
```

1st1:
Thanks for the PR! BTW, uvloop requires libuv @v1.9.0.  It won't compile with an earlier version.  Is there any way to check the version of libuv in `setup.py`, or we should just let the compiler to fail?

setup.py
mtorromeo: Snippet added but I cannot test it since I don't use Mac OS X.

1st1: I've tested it ;)

mtorromeo:
I'm not sure how feasible it would be to test libuv version but I'd rather keep things simple.

The shared object version would be a way to check for ABI compatibility but version 1.8.0 installed `libuv.so.1.0.0` just as 1.9.0 does.

Having the compiler just fail doesn't seem too bad to me as long as it is documented somewhere since `--use-system-libuv` is explicit and you should know what you are doing. Distribution packagers know how to check these things.

1st1:
Alright, merged in d65265689baf9b7a51952e980ec870df4a3a660b. Thanks a lot!

1st1:
FYI `v0.4.16` has this PR in.

PR #27, opened 2016-05-16
Build manylinux1_x86_64 wheel

methane:
FWIW, this pull request build manylinux1 wheel in Travis using docker.

FYI, cython provides manylinux1 wheel.  When using pip 8.1 or newer, installing cython is ultra fast without --no-binary option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,22 @@
 language: python
+sudo: required
 
 os:
     - linux
+
+services:
+    - docker
+
+env:
+    global:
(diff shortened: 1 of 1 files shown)


1st1:
It doesn't upload the wheels to PyPI, right?

methane:
Same to uvloop-release.tar.gz.  It uploads to github release when tagged in master branch.

1st1:
@elprans added the infrastructure for this (I think in a bit different way than this PR). Thanks a lot for giving it a shot!

PR #38, opened 2016-07-05
makefile: support the python binary being called something else

kaniini:
On FreeBSD ports, system python is usually python 2.7.

By modifying the makefile, we can override the python version used, by using `gmake PYTHON=python3.5` or otherwise appropriate.


1st1:
Merged by hand. Thanks a lot!

PR #46, opened 2016-08-13
Fix typo in create_datagram_endpoint

### LINKED ISSUE #45: UDPTransport isn't sending traffic to remote_addr when local_addr is set
* **uvloop version**: 0.5.2
* **python version**: Python 3.5.2
* **platform**: Archlinux

Using the [asyncio example udp echo server](https://github.com/python/asyncio/blob/master/examples/udp_echo.py) with uvloop. Works perfectly without uvloop, but once uvloop is used, the remote addr isn't set on the transport. Fails to send packets to the remote machine.

Note: It works if you run the server/client on the same machine. To replicate this, try going between two machines.

Inspecting the socket of the created transport, I see `<socket.socket fd=10, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('0.0.0.0', 55216)>`. 

With asyncio, raddr is set (`<socket.socket fd=6, family=AddressFamily.AF_INET, type=2050, proto=17, laddr=('10.200.0.9', 33074), raddr=('10.10.8.30', 9999)>`).

I'd like to use this with [aiosip](https://github.com/Eyepea/aiosip), but I'm also seeing this problem there.

vodik:
Typo causes `create_datagram_endpoint` to send traffic to the `local_addr` instead of the `remote_addr`.

Closes #45.

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -2316,7 +2316,7 @@ cdef class Loop:
                         local_addr)) from ctx
 
                 if rads is not None:
-                    rai = (<AddrInfo>lads).data
+                    rai = (<AddrInfo>rads).data
                     sock = udp._get_socket()
                     while rai is not NULL:
                         if rai.ai_family != lai.ai_family:


1st1:
Hi! Thank you for this PR! Would you be able to write a regression test?

vodik:
I'll see what I can do with different loopback address.

vodik:
Okay, I've modified the UDP test case to set a fixed local_addr of `127.0.0.2:9999`. This causes the UDP tests to fail on master, but work with this patch.

If you rather a separate test case, I can do that too, but mind giving me some pointers how I should organise things?

1st1:
@vodik With this PR `test_create_datagram_endpoint_addrs` fails on MacOS X now:

```
======================================================================
ERROR: test_create_datagram_endpoint_addrs (test_udp.Test_UV_UDP) (local_addr=('127.0.0.1', 0))
----------------------------------------------------------------------
  File "uvloop/loop.pyx", line 2304, in uvloop.loop.Loop.create_datagram_endpoint (uvloop/loop.c:36698)
    udp._bind(lai.ai_addr, reuse_address)
  File "uvloop/handles/udp.pyx", line 117, in uvloop.loop.UDPTransport._bind (uvloop/loop.c:88034)
    raise exc
OSError: [Errno 49] Can't assign requested address

The above exception was the direct cause of the following exception:

  File "/Users/yury/dev/magic/uvloop/tests/test_udp.py", line 77, in test_create_datagram_endpoint_addrs
    transport, client = self.loop.run_until_complete(coro)
  File "uvloop/loop.pyx", line 1133, in uvloop.loop.Loop.run_until_complete (uvloop/loop.c:19943)
    return future.result()
  File "uvloop/future.pyx", line 123, in uvloop.loop.BaseFuture.result (uvloop/loop.c:94147)
    return self._result_impl()
  File "uvloop/future.pyx", line 78, in uvloop.loop.BaseFuture._result_impl (uvloop/loop.c:93686)
    raise self._exception
  File "uvloop/task.pyx", line 126, in uvloop.loop.BaseTask._fast_step (uvloop/loop.c:99430)
    result = meth(None)
  File "uvloop/loop.pyx", line 2315, in create_datagram_endpoint (uvloop/loop.c:36980)
    raise OSError('could not bind to local_addr {}'.format(
OSError: could not bind to local_addr ('127.0.0.2', 0)
```

Would it work if we bind `local_addr` to `127.0.0.1` and some unused port?

vodik:
Yeah, that works. Actually, there's no reason to use a second address so long as we bind to port 0.

1st1:
`test_create_datagram_endpoint_addrs` still fails for me on MacOS, now with a `TimeoutError`.

Alright, could you please remove the unittest part from your PR and I'll merge it then. I'll try to figure out a way to test this properly later.

vodik:
No problem. I don't have a mac so I can't lend a hand. Figuring out something more proper might benefit the TCP tests too.

Maybe its as simple as putting them as a different test and documenting that 127.0.0.2 needs to be added to loopback. My understanding, after googling around, is that it works on mac. Just the networking stack defaults to 127.0.0.1/32 instead of 127.0.0.1/8.

1st1:
Merged, thanks a lot!

1st1:
Releasing 0.5.3 right now.

PR #53, opened 2016-09-11
setup.py: add missing space after '-fPIC' for ARCHFLAGS

zchee:
I have problem when install uvloop.

Basically, if users want to set `ARCHFLAGS`, maybe run the
```bash
ARCHFLAGS='-arch x86_64' python setup.py install
```
but it will define of `CFLAGS` to
```bash
CFLAGS=' -O2 -fPIC-arch x86_64'
```

It needs one space after `-fPIC`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ class libuv_build_ext(build_ext):
         if not re.search('-O\d', cur_cflags):
             cur_cflags += ' -O2'
 
-        env['CFLAGS'] = (cur_cflags + ' -fPIC' + env.get('ARCHFLAGS', ''))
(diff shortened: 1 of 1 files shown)


zchee:
Just in case, I know now `setup.py` can build with `ARCHFLAGS=' -arch x86_64'`(add space to before -arch), but I think this fix maybe more user-friendly.

asvetlov:
At least it doesn't brings any harm.
@1st1 I think you may just merge it.

1st1:
Thanks! Will merge in a couple of days.

1st1:
Thank you, this PR is merged now! I'll release a new version soon.

PR #58, opened 2016-10-24
Fixes broken image on PyPi entry

mikeputnam:
Preserves Github image rendering and fixes PyPi broken image.

Presuming this same README.rst is used verbatim for both Github and PyPi.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,7 @@ Performance
 
 uvloop makes asyncio 2-4x faster.
 
-.. image:: performance.png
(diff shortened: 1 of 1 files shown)


1st1:
Yes, that would fix the image to be displayed, however, PyPI doesn't try to resize the image to fit to the screen. Since that PNG file is hi-res, it breaks the PyPI page. I've played with various ReST parameters to adjust how `.. image::` is rendered, nothing really helps.

mikeputnam:
Hmm... I see.  Closing!

PR #87, opened 2017-04-10
Link with pthread

frederikaalund:
When importing uvloop, I got the following error:

    >>> import uvloop
      File "<stdin>", line 1, in <module>
      File "/projects/RedPitaya/OS/buildroot/buildroot-2017.02/output/target/usr/lib/python3.5/site-packages/uvloop/__init__.py", line 7, in <module>
    ImportError: /usr/lib/python3.5/site-packages/uvloop/loop.cpython-35m-arm-linux-gnueabihf.so: undefined symbol: pthread_atfork

This is fixed by linking the plugin (`loop.cpython-35m-arm-linux-gnueabihf.so`) to pthread as done in this patch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -267,6 +267,8 @@ class uvloop_build_ext(build_ext):
             self.compiler.add_library('kvm')
         elif sys.platform.startswith('sunos'):
             self.compiler.add_library('kstat')
+            
+        self.compiler.add_library('pthread')
 
         super().build_extensions()
 


1st1:
Do we need to do this for all platforms or for Linux+arm only?

frederikaalund:
> Do we need to do this for all platforms or for Linux+arm only?

Good point. Theoretically, it's needed for all operating systems that use POSIX threads, which is most Unix-like operating systems (Linux, MacOS, etc.). Practically speaking, that is every platform except Windows.

I'm surprised that I'm the first person to get the `undefined symbol` error. Apparently, this error only triggers in the Linux + ARM combination? Anyhow, linking explicitly against pthread is the way to go.

Also, consider using [`-pthread` instead of `-lpthread`](http://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling). I know that I use `-lpthread` in my patch; I learnt about the difference after making the pull request.

1st1:
Merged as is. Thanks!

PR #106, opened 2017-09-04
move remove_done_callback to finally block to fix RuntimeError('Event loop stopped before Future completed')

jimmylai:
@1st1 
This fix is the same as https://github.com/python/cpython/pull/1688

I still don't have clear idea about how the future with callback was left over in the loop.

This fix has proven to be effective in large scale production.


1st1:
Thanks @jimmylai

Issue #104, opened 2017-09-01
Difficult to cross-compile (e.g. Yocto)


cmcqueen:
* **uvloop version**: 0.8.0
* **Python version**: 3.5.2
* **Platform**: Cross-compile build using Yocto for BeagleBone Black (ARM)
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: N/A

I am trying to make a Yocto bitbake recipe to build uvloop. At the `build_ext` step it gives an error for the `libuv` `./configure` call, saying for cross-compilation it needs the `--host` parameter specified.

It might be possible to use the `--use-system-libuv` option, but at the moment Yocto doesn't support custom options for the `build_ext` step.

I applied the following patch to work around this:

    From 56604f7720089da7d9afd43825fc29f006a0e4bf Mon Sep 17 00:00:00 2001
    From: Craig McQueen <craig.mcqueen@innerrange.com>
    Date: Thu, 31 Aug 2017 16:31:17 +1000
    Subject: [PATCH] Allow for cross-compile of libuv
    
    If HOST_SYS environment variable is defined, then call ./configure with
    --host parameter set to HOST_SYS value.
    ---
     setup.py | 6 +++++-
     1 file changed, 5 insertions(+), 1 deletion(-)
    
    diff --git a/setup.py b/setup.py
    index ae9a65a..ad474e8 100644
    --- a/setup.py
    +++ b/setup.py
    @@ -226,8 +226,12 @@ class uvloop_build_ext(build_ext):
                  'Makefile.am', 'Makefile.in'],
                 cwd=LIBUV_BUILD_DIR, env=env, check=True)
     
    +        if 'HOST_SYS' in env:
    +            cmd = ['./configure', '--host=' + env['HOST_SYS']]
    +        else:
    +            cmd = ['./configure']
             subprocess.run(
    -            ['./configure'],
    +            cmd,
                 cwd=LIBUV_BUILD_DIR, env=env, check=True)
     
             j_flag = '-j{}'.format(os.cpu_count() or 1)
    -- 
    2.14.1

But then I had another problem, which is when I tried to import `uvloop`, I got an error regarding an unresolved symbol `pthread_atfork`. I used the following patch to resolve this:

    From abb18a622d5c3c341fd64da46b243d2734cbe2c8 Mon Sep 17 00:00:00 2001
    From: Craig McQueen <craig.mcqueen@innerrange.com>
[...]

PR #114, opened 2017-11-12
Allow for cross-compile of libuv

cmcqueen:
If LIBUV_CONFIGURE_HOST environment variable is defined, then call
./configure with --host parameter set to LIBUV_CONFIGURE_HOST value.


cmcqueen:
This is for issue #104.

1st1:
Thanks!

PR #138, opened 2018-04-10
Support for PY37 without contextvars

pfreixes:
This MR aligns the `uvloop` code to become compatible, not production
ready, with PY37 adding the following changes:

- Support for the new API used behind the `set_debug` function.
- Adds the `context=None` KW to those functions that will receive it
  when `uvloop` runs with PY37.
- Former asyncio.test_utils moved as an `uvloop` test resource.

New MR on top of that to enable `contextvars` should be expected.


tests/test_unix.py
```diff
@@ -362,6 +363,7 @@ def connection_lost(self, exc):
         self.assertIn(excs[0].__class__,
                       (BrokenPipeError, ConnectionResetError))
 
+    @unittest.skipUnless(sys.version_info < (3, 7), 'Python version must be < 3.7')
```
pfreixes: If you have some light about why this testing is failing with `3.7` please tell me.

uvloop/includes/stdlib.pxi
```diff
@@ -146,8 +145,8 @@ cdef py_inf = float('inf')
 
 
 # Cython doesn't clean-up imported objects properly in Py3 mode,
-# so we delete refs to all modules manually (except sys)
-del asyncio, concurrent, collections, errno
+# so we delete refs to all modules manually (except sys, asyncio)
+del concurrent, collections, errno
```
pfreixes: Just removed the `asyncio` to avoid losing the ability to call later the `asyncio.coroutines.debug_wrapper` when `uvloop` run on an interpreter different than 3.7. 

If you believe that this is a red flag, let's find another way to make the `_set_coroutine_debug ` ready to be executed by different Python versions.

pfreixes:
No way, Travis still failing one test gets frozen forever. 

My gut feeling says that is because the older 3.7-dev CPython version pointing still to `3.7.0a4+`. Tried to run the Travis step using Ubuntu 16.04 that must come with `3.7.0b1` [1] but looks like Travis does not have official and production-ready support for `Xenial` - WAT! - and keeps installing the `Trusty` version. 

Any ideas?

[1] https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/16.04/x86_64/python-3.7-dev.tar.bz2

1st1:
Yep, I saw this PR yesterday. Will take a look tomorrow. Thanks!

.ci/travis-install.sh
```diff
@@ -22,6 +22,7 @@ if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
     brew outdated automake || brew upgrade automake --with-default-names
 fi
 
-pip install --upgrade pip wheel
+#pip install --upgrade pip
```
pfreixes: Getting this error with the Mac environment

```bash
You should consider upgrading via the 'pip install --upgrade pip' command.
+pip install --upgrade setuptools
  File "/Users/travis/.pyenv/versions/3.6.3/bin/pip", line 7, in <module>
    from pip import main
ImportError: cannot import name 'main'
```

pfreixes:
Finally got a green CI, not really happy with some of the patches. in any case, its something that can be used to start and obviously to get improve it. Feedback welcomed @1st1.

I know that you had some plans to give support for `contextvars` - the natural continuation after this PR. if you are in a rush with other stuff, I can make it.

tests/test_udp.py
```diff
@@ -3,7 +3,7 @@
 import unittest
 import sys
 
-from asyncio import test_utils
+from tests import utils as test_utils
```
1st1: I don't like this "tests" import. Just move/copy the necessary functions to `uvloop._testbase`.

uvloop/includes/stdlib.pxi
1st1: We need to keep `del asyncio`. Otherwise uvloop would hold refs to Tasks longer than necessary (that's Cython's fault and these `del`s are a workaround).

You can use `getattr`:

```cython
cdef aio_debug_wrapper = getattr(asyncio.coroutines, 'debug_wrapper', None)
```

And just check if it's not ``None`` before you call it.

1st1:
> if you are in a rush with other stuff, I can make it.

I think I'll work on it myself. My idea is to copy `_asynciomodule.c` to uvloop. Since it's my code I think it shouldn't take me too long to figure out how to do that correctly.

1st1:
This PR needs to land before I start working on contextvars though.

pfreixes:
@1st1 Changes applied, about the issues with Pip and Python 3.7:

Travis CI uses an older `3.7-dev` image, exactly `3.7.0a4+`. And the `3.7` is no longer available. Since this breaking change [1] I would say that Python 3.7 is kinda broken for Travis. I hope that the Docker solution will be a temporary one, but right now is the only way to get a functional environment running an interpreter with a version beyond the alpha 4.

The Pip issue started with the recent 10.0.0 release and the Pyenv environments [2] [3]. Take a look to the last commit, just pins the PIP version for the Mac osx environment. I hope so again that this will be just a temporary solution.

[1] https://github.com/travis-ci/travis-ci/issues/9069
[2] https://github.com/pypa/pip/issues/5240
[3] https://github.com/pyenv/pyenv/issues/1141

1st1:
> Travis CI uses an older 3.7-dev image, exactly 3.7.0a4+. And the 3.7 is no longer available. Since this breaking change [1] I would say that Python 3.7 is kinda broken for Travis. I hope that the Docker solution will be a temporary one, but right now is the only way to get a functional environment running an interpreter with a version beyond the alpha 4.

Can you try installing python 3.7 using pyenv? It worked fine for me when we tried it for https://github.com/azure/azure-functions-python-worker/

pfreixes:
I didn't manage to make it run, but I'm still on it.

Brief summary. Initially I was trying to pull through `pipenv` the `3.7-dev` Python version, but most of the times Travis failed with a timeout because of the downloading process from Git plus the compilation process. So, later Ive tried pinning to the `3.7.0 beta 3` that is the last available by downloading it through FTP, and more important in my machines passes all of the tests with that version. Surprisingly, pinning to that version the timeouts during the install process in CI vanished.

Unfortunately, it turns out that the tests get frozen, exactly in the `test_create_connection_3` [1]. This is something that I've already experienced with alpha versions of the 3.7 - also in my local machine - but never with the beta ones. Indeed, as I said in my local machine all the tests work like a charm using the `3.7.0 beta 3` and the same using Docker.

So what is happening? No  idea. I've checked that the python version used to run the test is the right one installed by `pipenv` [2], and it seems to be right. 

I will try to work more on this,  first I need to reproduce the issue In my local machine otherwise it is gonna be really painful :)

Any suggestions will be wellcomed.

PD: The time needed to pass the Travis step using Docker for the Python 3.7 is more or less the same that is needed for the none dockerized steps. So, if you consider that this technical debt can be removed at some point - perhaps having the proper resolution by the Travis guys, we could think moe ahead with the Docker solution thinking on it as a temporary one. Agree, that if this is a testing issue has to be solved but my gut feeling says that this is not test issue, you can pull the branch and test it against your local Python 3.7 installation

PD2: Out of topic, I have a pending conversation with you regarding the tracing. I would like to give you another point of view. But I will work on it this weekend.

[1] https://github.com/MagicStack/uvloop/blob/master/tests/test_tcp.py#L424 
[2] https://github.com/MagicStack/uvloop/pull/138/files#diff-bd3c3a6aaea54a8f42fa6e7850a7a252R17

pfreixes:
@1st1 I didn't manage to reproduce the same issue in local. These have been the environments that ran the tests successfully in local:

- Docker image with Ubuntu 14.04 (trusty) with a OpenSSL 1.1 and Python 3.7 b03 from Pyenv,
- Docker image with Ubuntu 16.04 (xenial), with Python 3.7 b03 from Pyenv.
- Docker image with Ubuntu 16.04 (xenial), with Python 3.7 b04 (used the same one that is used by Travis)

I've tried to reproduce the local environment following this [guide](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally) but looks like that the Xenial is still experimental and there is no way to download the image used by Travis, or at least I wasn't able to find it. In any case, I've opened an [issue](https://github.com/travis-ci/travis-ci/issues/9577) to gather some feedback from Travis

Also, I've tried to get some information running the Travis CI, no luck and also the same issue: stuck in the same test `test_create_connection_3 (test_tcp.Test_AIO_TCP)`. Worth mentioning that the `xenial` environment is quite slower than the `trusty` one, usually takes 10 minutes more to start! and even sometimes the job finishes while the worker is still compiling the package (timeout?).

So, let's wait for some feedback from the Travis team. But I wouldn't put so many hopes on it, taking into account that for them the Xenial is still [experimental](https://github.com/travis-ci/travis-ci/issues/9038#issuecomment-356811024) and not officially supported by Travis.

pfreixes:
Definitely, [something](https://github.com/travis-ci/travis-ci/issues/9577#issuecomment-386885554) is wrong with Xenial and Travis

wumpus:
I am using pyenv, python 3.7.0b4 on CentOS 7, and installed this branch -- success!

It would be nice for this to get released so more people could test 3.7-dev. I personally also use Travis/Xenial for CI so the remaining bug affects me, but ...

pfreixes:
My proposal here is going for the Docker option that was proven as successful and TBH didn't add so much overhead in terms of time needed to run the tests.

Later if the `xenial` distro is officially supported by Travis and the tests can be run natively without using Docker its a matter of changing the configuration.

wumpus:
I think we are agreeing that uvloop should not wait for TravisCI.

1st1:
Closing this one in favour of #149. The new PR is a version of this one with removed CI tweaks.

pfreixes:
Thanks!

Issue #163, opened 2018-06-01
Implement tracing support


1st1:
This is a meta issue to discuss how new asyncio tracing API should look like. We will implement it first in uvloop, and later, if necessary, port it to asyncio.

-----------

#### Requirements

1. Base on the new `contextvars` (PEP 567) API, so that async tasks and callbacks they schedule are traced consistently.

2. Traces should be hierarchical, i.e. it should be possible to see what and how exactly one Task did, what Tasks that it spawned did, etc.

3. Low performance overhead. Ideally we should be able to run tracing in production.

4. Low-level and detailed: we want to be able to measure DNS requests, read/write callbacks performance, throughput, number of callbacks/tasks created etc.

5. Flexibility: traces should be in easy to work with format, so that it's possible to submit them to existing systems like zipkin.

6. Extensibility: the new API should be easy to maintain on both ends: in the event loop and in applications that use it.  This means that adding a new kind of trace should not require people to write any glue code.

7. [TBD]

1st1:
https://github.com/MagicStack/uvloop/pull/160 is the first attempt at implementing the new API.  

The main thing I don't like about it is that it's designed around having various `_begin` and `_end` callbacks in so called `Collector` objects.  There are a few problems with this approach:

* This isn't flexible: you'll need to update your collector with new `_begin` and `_end` methods whenever a new version of uvloop with new tracing bits comes out.

* It's too generic. I don't like that users will be able to run arbitrary code at arbitrary places. Writing a proper and low-overhead collector will be a non-trivial task. Speaking from my experience, we will end up in a situation where people do all kinds of crazy stuff in those callbacks, including scheduling more callbacks and doing disk IO.

In my opinion it would be better to create a `Span` dataclass that has the following attributes:

* `name: str` -- name of the trace, like "task.created", or "network.data_received".
* `id: uint64` -- unique span ID
* `parent_id: uint64` -- ID of the parent span
* `type: uint8`:
  - `0`: duration
  - `1`: counter

For "duration" spans:
* `created_at: uint64` -- local monotonic time (in nanoseconds?) when the span was created
* `duration: uint64` -- how long the span was active

For "counter" spans:
* `count: uint64` -- how many times a certain thing happened, or how many bytes were transferred

This is just an idea we should play with. The design is inspired by statsd and zipkin.

pfreixes:
Thanks for your feedback @1st1 

[...]

PR #171, opened 2018-06-21
POC Tracing support

pfreixes:
Implements a new tracing support for uvloop adding the pair methods
`start_tracing` and `stop_tracing` that allows the user to start or
stop the tracing. The user must provide a valid `uvloop.tracing.Tracer`
implementation that would be used internally by uvloop to create spans,
each span must also meet the `uvloop.tracing.Span` class contract.

This POC only implements the creation of spans during underlying calls
to the `getaddrinfo` function.

This POC relates to the conversation in #163.

What is missing/What is pending to discuss:

- `uvloop.tracing.Tracer.current_span` needs to be adapted to avoid race conditions, when a span is created the "global" attribute is overwritten. Most likely `current_span` will end up as another context variable.

- Agreement of an almost final draft of the `start_tracing` entry point, right now the implementation hides the `TracedContext`, so making it usable only internally by uvloop. Im wondering if we must make it public providing to the user a way of creating spans in his code, so becoming the way of trace ~asyncio~ uvloop code.

- Current implemetation implicitly wires the Tracer context using the parent span, so for example when internally a new span is created suing the `TracedContext.start_span` under the hood the parent span is passed to the `Tracer.create_span`, so allowing to the tracer to fetch some context information from this, also allowing make the new span child of the the parent one. Should we reconsidere this witih a more explicit interface ?

- Support for other spans within uvloop. But before doing this, we would need a kind of agreement of behind what circumstances the spans are created. A good example is the `create_connection` loop function, would be feel confortable with somehting like this:

```python:
async def create_connection(...)
    with __traced_context() as span:
        .....
```

That it would mean that when there is no tracing enabled it will create a noop span, the benefits of this global wrapping is making the code cleaner and easier to mantain but the drawbacks are quite clear.

Also worth mentioning that going to a pattern that uses a global context would mean that we will be creating a span when indeed in some failure scenarios we connected to nothing and without yielding the loop.

### DIFF
--- a/uvloop/__init__.py
+++ b/uvloop/__init__.py
@@ -1,15 +1,21 @@
 import asyncio
 
 from asyncio.events import BaseDefaultEventLoopPolicy as __BasePolicy
+from sys import version_info
 
 from . import includes as __includes  # NOQA
 from . import _patch  # NOQA
 from .loop import Loop as __BaseLoop  # NOQA
 
+PY37 = version_info >= (3, 7, 0)
+
+if PY37:
+    from .loop import start_tracing, stop_tracing
+    __all__ = ('new_event_loop', 'EventLoopPolicy', 'start_tracing', 'stop_tracing')
+else:
+    __all__ = ('new_event_loop', 'EventLoopPolicy')
 
 __version__ = '0.11.0.dev0'
-__all__ = ('new_event_loop', 'EventLoopPolicy')
-
 
 class Loop(__BaseLoop, asyncio.AbstractEventLoop):
     pass
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -18,7 +18,9 @@ from .includes.python cimport PY_VERSION_HEX, \
                               PyContext, \
                               PyContext_CopyCurrent, \
                               PyContext_Enter, \
-                              PyContext_Exit
+                              PyContext_Exit, \
+			      PyContextVar, \
+			      PyContextVar_Get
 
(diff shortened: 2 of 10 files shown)


uvloop/__init__.py
```diff
@@ -1,15 +1,21 @@
...
 from . import _patch  # NOQA
 from .loop import Loop as __BaseLoop  # NOQA
 
+PY37 = version_info >= (3, 7, 0)
 
-__version__ = '0.11.0.dev0'
-__all__ = ('new_event_loop', 'EventLoopPolicy')
+if PY37:
+    from .loop import start_tracing, stop_tracing
+    __all__ = ('new_event_loop', 'EventLoopPolicy', 'start_tracing', 'stop_tracing')
```
1st1: They should always be available, just raise `NotImplementedError` in 3.6/3.5.

uvloop/loop.pyx
```diff
@@ -821,13 +823,26 @@ cdef class Loop:
...
                 else:
                     if not fut.cancelled():
                         fut.set_result(data)
+
             else:
                 if not fut.cancelled():
                     fut.set_exception(result)
 
+            traced_context = __traced_context()
+            if traced_context:
```
1st1: Please use explicit `is None` checks.

1st1:
Looks good in general. I'll try to find time to get back to the discussion we have in the issue to see if we're missing anything.

1st1:
> What is missing/What is pending to discuss:

This is a great summary, please update the issue with it.

PR #174, opened 2018-06-26
 Add _run_once to be consistent with asyncio

frmdstryr:
Allows the user to manually run an iteration the event loop. Named to be consistent with https://github.com/python/cpython/blob/3.6/Lib/asyncio/base_events.py#L1338.

This allows you to "wait" for a future to complete within a sync function and is also useful for tests.

### DIFF
--- a/tests/test_sockets.py
+++ b/tests/test_sockets.py
@@ -189,6 +189,91 @@ class _TestSockets:
             self.assertEqual(sock.fileno(), -1)
             self.loop.run_until_complete(asyncio.sleep(0.01, loop=self.loop))
 
+    def test_sock_cancel_add_reader_race(self):
+        srv_sock_conn = None
+
+        async def server():
+            nonlocal srv_sock_conn
+            sock_server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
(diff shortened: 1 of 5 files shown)


1st1:
I don't like the idea of supporting private asyncio APIs and implementation details in uvloop. `_run_once()` is one of such methods that we might get rid of in asyncio in Python 3.8 too. Therefore my inclination right now is to close this as "won't fix".

A simple workaround that uses only public APIs is to call `loop.run_until_complete(asyncio.sleep(0))`. Would that work for you?

frmdstryr:
`loop.run_until_complete` doesn't work if the loop is already running.

Perhaps just changing `_on_idle` to a `cpdef` would be acceptable?

1st1:
> loop.run_until_complete doesn't work if the loop is already running.

If the loop is already running you can use `loop.call_soon(call_me_after_one_loop_iteration)` or `await sleep(0)`.

> Perhaps just changing _on_idle to a cpdef would be acceptable?

No, not really, as people will start depending on it and that would make it harder to refactor uvloop in the long term.

frmdstryr:
Ahh, ok.  Thanks for the prompt replies.

PR #176, opened 2018-07-02
[WIP] Rewrite sslproto.py

fantix:
Refs #158, this is a WIP yet, please kindly wait until ready for review.


1st1:
> According to libuv docs, the following read_cb might be called with nread=UV_EOF with no buffer allocated.

Ah, that makes sense. Thanks for fixing that; can you make a fix in a separate PR?

What's the projected timeline for this PR? I assume you are planning to add more SSL tests, right? Or is it ready for a review now?

fantix:
> Ah, that makes sense. Thanks for fixing that; can you make a fix in a separate PR?

Oh sure thing, will do.

> What's the projected timeline for this PR? I assume you are planning to add more SSL tests, right? Or is it ready for a review now?

Yes, I was planning to add more tests (renegotiation for example), and fix some obvious issues and make improvements during the testing. Let's do the review tomorrow anyway, shall we? 😃

1st1:
Anytime this week works.

tests/test_tcp.py
```diff
@@ -9,6 +9,7 @@
 import threading
 import weakref
 
+from OpenSSL import SSL
```
1st1: Maybe `from OpenSSL import SSL as open_ssl`? :)  To make it obvious where OpenSSL is really used.

tests/test_tcp.py
```diff
@@ -1255,6 +1256,144 @@ def prog(sock):
...
+
+        clients = []
+
+        async def handle_client(reader, writer):
+            nonlocal CNT
+
+            # hack reader and writer to call start_tls()
+            transport = writer._transport
+            writer._transport = None
+            reader._transport = None
```
1st1: Wow, you're brave! :)

tests/test_tcp.py
fantix: Sure

tests/test_tcp.py
fantix: LOL was rushing to avoid manual protocol code - perhaps it's better be replaced by protocol so that the test can be robust enough even if asyncio changes its implementation in the future.

uvloop/sslproto.pyx
```diff
@@ -471,62 +297,39 @@ class SSLProtocol(object):
...
-                except Exception as ex:
-                    self._fatal_error(
-                        ex, 'application protocol failed to receive SSL data')
-                    return
-            else:
-                self._start_shutdown()
-                break
+        elif self._state == _SHUTDOWN:
+            self._do_read()
+            self._do_shutdown()
```
fantix: `_do_read()` should be removed in `_SHUTDOWN`, because a peer starts to discard incoming app data once it sends the `close_notify` alert.

1st1:
In short: I like the new code, it's easy to follow. Great job! Don't have any major comments. The state transitions seem fine, the overall logic is OK too.

I'd continue iterating on the codebase and adding tests (your current ToDo is great too!)

At some point when we think it's ready we'll commit the Python implementation (to have it in the repo) and start to Cythonize it to make it slightly faster. The pure-Python implementation will go straight to asyncio 3.8.

fantix:
I'm adding a new state `FLUSHING` before `SHUTDOWN` (please see updated diagram at top), for cases that:

* `transport.close()` is called, but there is still data in `_write_backlog`
* `close_notify` is received, but there is still data in `_write_backlog`

Usually APP data will transiently stay in `_write_backlog` and immediately be encrypted into SSL data in `outgoing` buffer. But when the other peer starts an renegotiation, `SSLObject` will stop the encryption work and wait for renegotiation to finish. So both cases do exist.

While in `FLUSHING` state:

* No further data will be fed to APP protocol - it'll be simply discarded
* Writing to the transport has no effect - data is discarded and warning will be raised after a few attempts, similar to the behavior of `TCPTransport`
* The `SSLProtocol` keeps trying to flush the data in `_write_backlog`, and moves to `SHUTDOWN` state once done
* The newly-added `ssl_shutdown_timeout` also covers this state - that means the countdown starts when it enters `FLUSHING` state

fantix:
Working on `eof_received()` case when peer closes the writing channel only, we still flush all pending data if any if possible, send `close_notify` without waiting for peer's `close_notify` and close the transport.

Also making the call to APP's `connection_lost()` earlier, as soon as:

* `close_notify` is received
* `transport.close()` is called
* `eof_received()` is called

instead of waiting for the underlying transport closes and calls our `connection_lost()`, so that APP protocol always get a `connection_lost()` before `transport.write()` stops working, unless `transport.close()` is explicitly called right before `transport.write()`.

fantix:
Spent some time to make it run faster in uvloop - about 20% ~ 40% faster than original uvloop in 1~100 KiB echo benchmark. Here're the cProfiled diagrams for time spent on each module:

time

And here's the SSL benchmark running [modified vmbench](https://github.com/fantix/vmbench) on a 4-core 2.5GHz cloud server, all benchmarked servers are running in `python:3.7` (Debian 9) docker containers ([full benchmark data here](https://htmlpreview.github.io/?https://github.com/fantix/vmbench/blob/master/ssl-result.html)):

benchmark

Where:

* `ssl raw` is pure OpenSSL 1.1.0f echo encryption and decryption without network I/O written in C++ based on [openssl-bench](https://github.com/ctz/openssl-bench) initially [referenced by Elvis](https://jbp.io/2018/01/07/rustls-vs-openssl-performance-1.html), it is the theoretical computing limit.
* `golang` (1.7.4) implemented its own TLS transport, well optimized it is assumed, thus it is considered as the computing limit + network I/O.
* `ssl uvloop protocol` is using the new `sslproto.pyx` with regular `EchoProtocol` inherited from `asyncio.Protocol`.
* `ssl uvloop buffered` is similar, but uses `EchoBufferedProtocol` which has less memory copies, but additional calls added more overhead.
* `ssl orig uvloop` is original uvloop with `asyncio.Protocol`.
* `nodejs` is on version 4.8.2.
* `ssl asyncio protocol` is vanilla asyncio with `asyncio.Protocol`.
* `gevent` is on version 1.3.5 (with [greenlet 0.4.13](https://github.com/gevent/gevent/issues/1260)).
* `ssl threaded` is copied from Curio.
* `twisted` is on version 18.7.0.

The new `sslproto.pyx` is performing slightly better than Golang at 100KiB, but `MemoryBIO` added much memory copy overhead which is significant for smaller payloads, still it is 10% ~ 20% faster than original uvloop on throughtput, and 60%~80% faster than cpython.

uvloop/sslproto.pxd
```diff
@@ -0,0 +1,113 @@
+cdef enum ProtocolState:
```
1st1: I'd call it `SSLProtocolState`. And maybe we don't need the leading `_`?

uvloop/sslproto.pyx
```diff
@@ -699,34 +834,18 @@ class SSLProtocol(object):
...
-            self._transport.close()
-
-    def _abort(self):
-        try:
-            if self._transport is not None:
-                self._transport.abort()
-        finally:
-            self._finalize()
 
+cdef _add_water_defaults(high, low, kb):
```
1st1: Can you put this function elsewhere and call it from `UVBaseTransport._set_write_buffer_limits()`?

fantix: Sure thing!

uvloop/sslproto.pyx
```diff
@@ -277,13 +28,13 @@ class _SSLProtocolTransport(aio_FlowControlMixin, aio_Transport):
 
     def get_extra_info(self, name, default=None):
         """Get optional transport information."""
-        return self._ssl_protocol._get_extra_info(name, default)
+        return (<SSLProtocol>self._ssl_protocol)._get_extra_info(name, default)
```
1st1: Would it be a good idea to make this a `cdef class`? We don't need too many methods from `asyncio.Transport` (and working `isinstance(ssl_proto_transport, asyncio.Transport)` is not a requirement.

If we make this an extension class you won't need `<SSLProtocol>` casts.

uvloop/sslproto.pxd
fantix: Right, they were globals and I forgot to rename them.

uvloop/sslproto.pyx
```diff
@@ -626,63 +497,327 @@ class SSLProtocol(object):
...
+            self._write_buffer_size += len(data)
+
+        try:
+            if self._state == _WRAPPED:
+                self._do_write()
+
+        except Exception as ex:
+            self._fatal_error(ex, 'Fatal error on SSL protocol')
+
+    def _append_write_backlog(self, data):
```
1st1: Can we prefix it this `_test_`?

fantix: Sure 😃

uvloop/sslproto.pyx
fantix: Yeah I was thinking about that too, but didn't go that way because I found no `cdef` methods in transport, and it was checked somewhere that `FlowControlMixin` must be a parent class. Let me check and get back to you. If possible, I prefer `cdef class` too with less dict lookups. 😈

fantix:
Thanks a lot for reviewing in late night! ❤️

uvloop/sslproto.pyx
fantix: Got it, I'll take a quick look and include it if it's simple

fantix:
Totally reasonable. Will do, and I'll see if I can find some more tools or ways to check possible memory issues.

examples/bench/echoserver.py
```diff
@@ -74,6 +74,21 @@ def data_received(self, data):
         self.transport.write(data)
 
 
+class EchoBufferedProtocol(asyncio.BufferedProtocol):
+    def connection_made(self, transport):
+        self.transport = transport
+        self.buffer = bytearray(256 * 1024)
```
1st1:
```python
self._buffer = bytearray(256 * 1024)
self.buffer = memoryview(self._buffer)
```

fantix: Ha! Been there, should have commented - to make sure the echo is true, the buffer needs to be copied, so that next read doesn’t change previous outgoing data which is still pending. With memoryview and bytes again, the benchmark gets slightly slower, therefore it ended up like this now. Thanks anyway! I’ll add comment.

1st1: Ah, right :)

We'll need to add new API to asyncio 3.8 to tell us when a `write()` has completed so that we can re-use the buffer.

fantix: That’s interesting!

1st1:
> Check sendfile support

We don't support the new sendfile in uvloop yet.

1st1:
> And here's the SSL benchmark running modified vmbench on a 4-core 2.5GHz cloud server, all benchmarked servers are running in python:3.7 (Debian 9) docker containers (full benchmark data here):

BTW, we should merge your additions back to the vmbench project.

fantix:
I could reproduce the memory pattern. The memory increase seems to be more relevant to new SSL connection, rather than repeated echo send and recv. So I tested with `python examples/bench/echoclient.py --workers 6 --msize 10241 --num 10 --ssl --times 10000`, and memory usage of the server stabilized at around 30MB. Increasing the number of concurrent workers to 12, the server memory usage ended up at around 40MB.

I've also reviewed the code, and made a few small changes. (I'll create a separate PR for the vmbench changes)

If okay, I'll rebase again and squash all commits.

1st1:
> If okay, I'll rebase again and squash all commits.

go for it!

uvloop/includes/flowcontrol.pxd
```diff
@@ -0,0 +1,20 @@
+cdef inline add_water_defaults(high, low, int kb):
```
elprans: I'd call this `add_flowcontrol_defaults()`.

uvloop/sslproto.pyx
```diff
@@ -382,27 +177,56 @@ class _SSLProtocolTransport(aio_FlowControlMixin, aio_Transport):
...
 
-class SSLProtocol(object):
+cdef class SSLProtocol:
     """SSL protocol.
 
     Implementation of SSL on top of a socket using incoming and outgoing
     buffers which are ssl.MemoryBIO objects.
     """
 
+    def __cinit__(self, *_, **__):
```
elprans: We usually spell this `*args`, `**kwargs`, even if unused.

1st1:
@fantix hey, could you please take a look at the failed CI?

fantix:
Sure thing! Sorry been out of office, will check today.

fantix:
Seems to be an unstable test, a retest passed. Tried but didn't reproduce locally, I'll try to find clues in the failing log.

1st1:
@asvetlov could you also take a look?  This is something that we'll use pretty much as is in asyncio in 3.8.

asvetlov:
Huge improvement! Thanks!

PR #179, opened 2018-07-04
Add python 3.7 on CI

waghanza:
Hi,

This `PR` is an attempt to add CI on **python** `3.7`

- [x] Linux 
- [x] OSX
- [x] Windows

**3.7** is not available by default on travis / linux @see https://github.com/travis-ci/travis-ci/issues/9831

Windows is already using **3.7**

Regards,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,13 @@ matrix:
           env: BUILD=tests,wheels PYTHON_VERSION=3.6.5
           branches: {only: [releases]}
 
+        - os: osx
+          osx_image: xcode7.3
+          # Travis macOS env does not support Python yet,
+          # so we have to set things up manually in install.sh.
+          env: BUILD=tests,wheels PYTHON_VERSION=3.7.0
+          branches: {only: [releases]}
+
         - os: linux
           dist: trusty
           sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,17 @@ matrix:
           services: [docker]
           env: BUILD=tests,wheels,release
 
+        - os: linux
+          dist: xenial
+          branches: {only: [releases]}
+          sudo: required
+          language: python
+          python: "3.7"
+          services: [docker]
+          env: BUILD=tests,wheels,release
+
+cache:
+
 cache:
     pip
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: generic
 env:
     global:
         - PYMODULE=uvloop
-        - RELEASE_PYTHON_VERSIONS="3.5 3.6"
+        - RELEASE_PYTHON_VERSIONS="3.5 3.6 3.7"
 
         - S3_UPLOAD_USERNAME=oss-ci-bot
         - S3_UPLOAD_BUCKET=magicstack-oss-releases
@@ -80,8 +80,6 @@ matrix:
           python: "3.7"
           services: [docker]
           env: BUILD=tests,wheels,release
-
-cache:
 
 cache:
     pip


.travis.yml
```diff
@@ -65,6 +72,17 @@ matrix:
...
+        - os: linux
+          dist: xenial
+          branches: {only: [releases]}
+          sudo: required
+          language: python
+          python: "3.7"
+          services: [docker]
+          env: BUILD=tests,wheels,release
+
+cache:
```
1st1: This line seems out of place.

1st1:
While you're here, please update `RELEASE_PYTHON_VERSIONS="3.5 3.6"` to `RELEASE_PYTHON_VERSIONS="3.5 3.6 3.7"` in `.travis.yml`.

1st1:
@elprans please take a look.

.travis.yml
waghanza: :tada:

.travis.yml
```diff
@@ -65,6 +72,15 @@ matrix:
           services: [docker]
           env: BUILD=tests,wheels,release
 
+        - os: linux
```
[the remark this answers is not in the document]
waghanza: waiting https://travis-ci.org/MagicStack/uvloop/builds/400060588 to finish

may I suggest using **1** `CI` that support `OSX` / `Linux` / `Windows` (instead of **2**)
I've found https://cirrus-ci.org/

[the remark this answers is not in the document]
1st1: Seems that Cirrus CI is just a couple of days old. We'll wait until it's mature before spending any time trying to integrate it.

.travis.yml
```diff
@@ -65,6 +72,15 @@ matrix:
...
+          dist: xenial
+          branches: {only: [releases]}
+          sudo: required
+          language: python
+          python: "3.7"
+          services: [docker]
+          env: BUILD=tests,wheels,release
```
1st1: Drop "wheels,release" from `BUILD`.

.travis.yml
```diff
@@ -65,6 +72,15 @@ matrix:
           services: [docker]
           env: BUILD=tests,wheels,release
 
+        - os: linux
+          dist: xenial
+          branches: {only: [releases]}
```
1st1: Drop `branches`

.travis.yml
```diff
@@ -65,6 +72,15 @@ matrix:
...
+          sudo: required
+          language: python
+          python: "3.7"
+          services: [docker]
```
1st1: Drop `services`

.travis.yml
waghanza: Could also be the same for other python versions, no ?

1st1: It's the same for other versions. There's only one target with `BUILD=tests,wheels,release` (3.5) which builds the wheels and releases them.  All other targets (except macos) only run tests.  So 3.7 target should also just run tests.

.travis.yml
```diff
@@ -42,28 +42,31 @@ matrix:
...
         - os: linux
-          dist: trusty
-          branches: {only: [releases]}
+          dist: xenial
           sudo: required
           language: python
-          python: "3.5"
-          services: [docker]
-          env: BUILD=tests,wheels,release
+          python: "3.7"
```
1st1: Eh, we needed to keep `env: BUILD=tests`. Never mind, I'll make a PR myself.

waghanza: :+1: reverting my change, it will be better to refactor test after this

1st1:
I pushed a CI change in a separate commit: https://github.com/MagicStack/uvloop/commit/37f964b757e48ae89ba7471df40345dc0d0fb233

Thank you for working on this, I credited you for the change!

waghanza:
@1st1 awesome :tada: but why removing **appveyor** ?

1st1:
AppVeyor was testing linux+python 3.7 (uvloop doesn't support Windows yet), which is now unnecessary.

PR #191, opened 2018-08-07
Fix contextvars.Context.copy memory leak.

hellysmile:
Hey, after switching to 3.7 I've found that our servers being leaked according to their load.
Switching to asyncio loop removes this leak

here is pseudo code which I used to reproduce leak

```
import asyncio
import janus
import queue as q

from pympler import muppy
from pympler import summary


all_objects1 = muppy.get_objects()


def produce(queue):
    for x in range(1000):
        print(x)
        while True:
            try:
                queue.sync_q.put(x, timeout=0.1)
                break
            except q.Full:
                continue


async def consume(queue):
    while True:

        item = await queue.async_q.get()
        if item is ...:
            await queue.async_q.put(...)

            break

        await asyncio.sleep(0.001)


import uvloop
asyncio.set_event_loop(loop)

queue = janus.Queue(3)

coros = [
    loop.run_in_executor(None, produce, queue)
    for _ in range(10)
]

workers = set()

for _ in range(3):
    worker = loop.create_task(consume(queue))
    workers.add(worker)
    worker.add_done_callback(workers.remove)

loop.run_until_complete(asyncio.gather(*coros))

loop.run_until_complete(queue.async_q.put(...))

loop.run_until_complete(asyncio.gather(*workers))

queue.close()
loop.run_until_complete(queue.wait_closed())

all_objects2 = muppy.get_objects()

sum1 = summary.summarize(all_objects1)
sum2 = summary.summarize(all_objects2)

diff = summary.get_diff(sum1, sum2)

summary.print_(diff)
import ipdb; ipdb.set_trace()
loop.close()
```

`summary.print_` shows ~20k of unallocated contexts

This fix does remove this leak and all tests are passed

### DIFF
--- a/uvloop/cbhandles.pxd
+++ b/uvloop/cbhandles.pxd
@@ -1,6 +1,7 @@
 cdef class Handle:
     cdef:
         Loop loop
+        bint _context_copied
         object context
         bint _cancelled
 
@@ -28,6 +29,7 @@ cdef class TimerHandle:
         bint _cancelled
         UVTimer timer
         Loop loop
+        bint _context_copied
         object context
         object __weakref__
 
--- a/uvloop/cbhandles.pyx
+++ b/uvloop/cbhandles.pyx
@@ -5,8 +5,7 @@ cdef class Handle:
         self._cancelled = 0
         self.cb_type = 0
         self._source_traceback = None
-
-        self._context_copied = False
+        self._context_copied = 0
 
     cdef inline _set_loop(self, Loop loop):
         self.loop = loop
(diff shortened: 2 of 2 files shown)


hellysmile:
travis build for https://github.com/MagicStack/uvloop/pull/191/commits/042d360324c482b07788c3be0c3aced1764f61f4 failed seems due slow travis mac ssl connection and seems not related to this PR, cuz previous commit is green

Can someone restart these builds?

1st1:
Wow! Thank you **so much** for tracking this down!

I have a different idea how to fix this though. Instead of fixing refcount in `__dealloc__` we should do that the moment we copy the context. Otherwise we can get another kind of leak if there's a cycle between the context and the task/callback handle. Hence https://github.com/MagicStack/uvloop/pull/192.  I'll close this one.

I will issue a bugfix release for 0.11.x and 0.10.x today.

1st1:
v0.11.2 has just been released with a fix for this. Thanks again!

Issue #197, opened 2018-09-21
Segfault in test_start_tls_client_buf_proto_1


1st1:
New TLS implementation seems to crash on Mac OS X and Linux.

Steps to reproduce:

1. Fresh build of `master`.
2. Run `pytest`.

Output:

```
[...]
tests/test_tcp.py::Test_UV_TCPSSL::test_remote_shutdown_receives_trailing_data PASSED
tests/test_tcp.py::Test_UV_TCPSSL::test_renegotiation PASSED
tests/test_tcp.py::Test_UV_TCPSSL::test_shutdown_cleanly PASSED
tests/test_tcp.py::Test_UV_TCPSSL::test_shutdown_timeout PASSED
tests/test_tcp.py::Test_UV_TCPSSL::test_ssl_connect_accepted_socket PASSED
tests/test_tcp.py::Test_UV_TCPSSL::test_ssl_handshake_timeout PASSED
tests/test_tcp.py::Test_UV_TCPSSL::test_start_tls_client_buf_proto_1 fish: “pytest -v” terminated by signal SIGSEGV (Address boundary error)
```

@fantix please take a look.

---------

Screenshot of Mac OS crash report (just for the reference; stack dump is always random):

<img width="1027" alt="screen shot 2018-09-21 at 1 20 38 pm" src="https://user-images.githubusercontent.com/239003/45896195-6237ab80-bda1-11e8-8829-a787cffb41cf.png">

elprans:
For reference, here's how it crashes on my machine:

```
tests/test_tcp.py::Test_UV_TCPSSL::test_start_tls_client_buf_proto_1 
Fatal Python error: GC object already tracked

Thread 0x00007f65c27fc700 (most recent call first):
  File "/usr/lib64/python3.6/ssl.py", line 633 in read
  File "/usr/lib64/python3.6/ssl.py", line 871 in read
  File "/usr/lib64/python3.6/ssl.py", line 994 in recv
  File "/home/elvis/dev/magic/uvloop/uvloop/_testbase.py", line 348 in recv_all
  File "/home/elvis/dev/magic/uvloop/tests/test_tcp.py", line 1679 in serve
  File "/home/elvis/dev/magic/uvloop/uvloop/_testbase.py", line 486 in _handle_client
  File "/home/elvis/dev/magic/uvloop/uvloop/_testbase.py", line 477 in _run
  File "/home/elvis/dev/magic/uvloop/uvloop/_testbase.py", line 446 in run
[...]

PR #198, opened 2018-09-29
Fix sslproto crash and test issues

fantix:
#197 This PR fixed:

* Unstable `test_remote_shutdown_receives_trailing_data`
* Hanging `test_write_to_closed_transport` with OpenSSL 1.1.1
* To close SSL transport instead of TCP transport when `await waiter` raises in `loop.pyx`, which caused `test_ssl_handshake_timeout` to crash its following tests.

TODOs after this PR:

- [ ] Check why `test_create_server_buffered_1` fails
- [ ] Check why the crash happened (on OpenSSL 1.1.1 only, not with `make debug`, with `pytest` only, how to reproduce the duplicate freeing and how to prevent it)

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1560,16 +1560,17 @@ cdef class Loop:
         transport.set_protocol(ssl_protocol)
         conmade_cb = self.call_soon(ssl_protocol.connection_made, transport)
         resume_cb = self.call_soon(transport.resume_reading)
+        app_transport = (<SSLProtocol>ssl_protocol)._app_transport
 
         try:
             await waiter
         except Exception:
-            transport.close()
+            app_transport.close()
             conmade_cb.cancel()
             resume_cb.cancel()
             raise
 
-        return (<SSLProtocol>ssl_protocol)._app_transport
+        return app_transport
 
     @cython.iterable_coroutine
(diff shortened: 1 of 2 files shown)


uvloop/loop.pyx
```diff
@@ -1560,16 +1560,17 @@ cdef class Loop:
         transport.set_protocol(ssl_protocol)
         conmade_cb = self.call_soon(ssl_protocol.connection_made, transport)
         resume_cb = self.call_soon(transport.resume_reading)
+        app_transport = (<SSLProtocol>ssl_protocol)._app_transport
 
         try:
             await waiter
         except Exception:
-            transport.close()
+            app_transport.close()
```
1st1: Wow, we need to backport this fix to CPython. Care to submit a PR?

uvloop/loop.pyx
```diff
@@ -1937,12 +1939,13 @@ cdef class Loop:
             tr._attach_fileobj(sock)
 
         if ssl:
+            app_transport = (<SSLProtocol>protocol)._app_transport
             try:
                 await ssl_waiter
             except Exception:
-                tr._close()
+                app_transport.close()
```
1st1: ditto

uvloop/loop.pyx
fantix: No problem, will do!

PR #203, opened 2018-10-10
Fix segfault and make SSL transport creation lazy

### LINKED ISSUE #197: Segfault in test_start_tls_client_buf_proto_1

fantix:
Fixes #197 

- [x] Stop using `self` in `__dealloc__`
- [x] Add a test to cover this
- [x] Make SSL transport creation lazy to avoid false resource warning
- [x] Fix error log in `test_create_connection_ssl_failed_certificate`

### DIFF
--- a/uvloop/sslproto.pyx
+++ b/uvloop/sslproto.pyx
@@ -49,10 +49,10 @@ cdef class _SSLProtocolTransport:
 
     def __dealloc__(self):
         if not self._closed:
-            _warn_with_source(
-                "unclosed transport {!r}".format(self),
-                ResourceWarning, self)
             self._closed = True
+            warnings_warn(
+                "unclosed transport <uvloop.loop._SSLProtocolTransport "
+                "object>", ResourceWarning)
 
     def is_reading(self):
         return not self._ssl_protocol._app_reading_paused
@@ -249,7 +249,8 @@ cdef class SSLProtocol:
         self._waiter = waiter
         self._loop = loop
         self._set_app_protocol(app_protocol)
-        self._app_transport = _SSLProtocolTransport(self._loop, self)
+        self._app_transport = None
+        self._app_transport_created = False
         # transport, ex: SelectorSocketTransport
         self._transport = None
         self._call_connection_made = call_connection_made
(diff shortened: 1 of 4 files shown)


uvloop/sslproto.pyx
```diff
@@ -301,6 +301,11 @@ cdef class SSLProtocol:
                 self._waiter.set_result(None)
         self._waiter = None
 
+    def _get_app_transport(self):
+        if self._app_transport is None:
+            self._app_transport = _SSLProtocolTransport(self._loop, self)
```
1st1: We set `self._app_transport = None` in `connection_lost`.  I'm a bit worried that we might end up in a situation (accidentally or while debugging/refactoring) where we create `_SSLProtocolTransport` more than once for one protocol.  Thoughts?

fantix: Yeah I had the same concern. How about adding an additional flag to make sure it's created only once?

1st1: Yeah, feel free to add one.

PR #204, opened 2018-10-13
Don't raise "requires a DNS lookup" error on Unix Domain Socket

### LINKED ISSUE #196: UVLoop raises '
* **uvloop version**: 0.11.2
* **Python version**: 3.7
* **Platform**: Debian 9.5

When using uvloop with Unix Domain Sockets `uvloop.loop.UDPTransport.sendto` raises `ValueError: UDP.sendto(): address '(socket path)' requires a DNS lookup` which is wrong, as there are no DNS lookups for Unix Domain Sockets.

I think this can be related to https://github.com/MagicStack/uvloop/issues/91/

Code to reproduce:
```
import asyncio
import os
import random
import socket
import string

import uvloop


def create_sock(path):
    try:
        os.unlink(path)
    except FileNotFoundError:
        pass
    s = socket.socket(socket.AF_UNIX, type=socket.SOCK_DGRAM)
    s.bind(path)
    return s


async def main():
    # Get a reference to the event loop as we plan to use
    # low-level APIs.

    message = "Hello World!"
    filename = '/tmp/{}'.format(''.join(random.choice(string.ascii_letters) for i in range(30)))
    sock = create_sock(filename)

    transport, protocol = await loop.create_datagram_endpoint(
        lambda: asyncio.DatagramProtocol(),
        sock=sock)

    transport.sendto(message.encode(), filename)

asyncio.run(main())
```

pax0r:
This fixes #196


tests/test_udp.py
```diff
@@ -155,6 +158,25 @@ def test_create_datagram_endpoint_sock(self):
...
+        tmp_file = os.path.join(tempfile.gettempdir(),  str(uuid.uuid4()))
+        sock = socket.socket(socket.AF_UNIX, type=socket.SOCK_DGRAM)
+        sock.bind(tmp_file)
+
+        with sock:
+            try:
+                f = self.loop.create_datagram_endpoint(
+                    lambda: MyDatagramProto(loop=self.loop), sock=sock)
+            except TypeError as ex:
+                # asyncio in 3.5.0 doesn't have the 'sock' argument
```
1st1: Better to simply skip this test under 3.5.

1st1:
Released in uvloop v0.11.3.

PR #205, opened 2018-10-30
backlog should be passed to create_unix_server

hikoz:
`backlog` parameter is not only for tcp server.


1st1:
Thank you!

1st1:
Released in uvloop v0.11.3.

Issue #214, opened 2019-01-08
UDPTransport.sendto spends a lot of time validating the address


jlaine:
* **uvloop version**: 0.11.3 and 0.12.0rc2
* **Python version**: 3.7
* **Platform**: Linux

I have been trying to optimize the performance of `aiortc`, an asyncio-based WebRTC implementation. `aiortc` uses `aioice` (an Interactive Connectivity Establishment library) for its network access. Using `uvloop` provides some performance improvement, but not as much as I had hoped.

During my optimization work I noticed that when called with an address, UDPTransport.sendto spends a noticeable amount of time calling into Python code, looking up the socket's `family` and `type` properties. Looking at the code, this corresponds to validation checks on the destination address which are run on every single packet (which involves a call to getaddrinfo whose result is ultimately discarded):

https://github.com/MagicStack/uvloop/blob/df0e543a653cd17b800a866b7efa5326352efa16/uvloop/handles/udp.pyx#L146

I am aware that specifying a remote address when calling `create_datagram_endpoint` would probably work around this issue, but in the case of ICE this is not practical as the remote party may have multiple addresses, which are not known in advance and may change over time.

To assess the impact of the validation code I tried a basic patch which stores the last validated address and skips the validation if `sendto` is called for the same address again. I then ran the following test:

```python
import asyncio
import socket
import time

import uvloop


class DummyProtocol(asyncio.DatagramProtocol):
    pass


async def run(loop):
    addr = ('127.0.0.1', 1234)
    data = b'M' * 1500
    transport, protocol = await loop.create_datagram_endpoint(
        lambda: DummyProtocol(),
        family=socket.AF_INET)

    start = time.process_time()
    for i in range(1000000):
        transport.sendto(data, addr)
    print('elapsed', time.process_time() - start)


asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
loop = asyncio.get_event_loop()
loop.run_until_complete(run(loop))
```

On my laptop the results of sending out a million packets:

- without the patch: 7.8s
- with the patch: 3.0s

Questions:
- do we need the validation code at all?
- if we do, would you consider a patch which caches the last successfully validated address and skips the validation if `sendto` is called again with the same address?
- alternatively, could we at the very least store the socket type and family to avoid the call into Python to fetch these two properties?

jlaine:
The trivial patch I used:

```patch
diff --git a/uvloop/handles/udp.pxd b/uvloop/handles/udp.pxd
[...]

PR #215, opened 2019-01-08
Improve UDPTransport.sendto address validation (fixes: #214)

jlaine:
When UDPTransport.sendto is called with a destination address, only
perform the validation once and store the result in an LRU cache. Also
store some of the socket's properties (family, proto, type) as integers
to avoid repeatedly fetching these properties and converting them to integers.

### DIFF
--- a/uvloop/handles/udp.pyx
+++ b/uvloop/handles/udp.pyx
@@ -1,4 +1,19 @@
-import socket
+import functools
+
+
+@functools.lru_cache()
+def validate_address(object addr, int sock_family, int sock_type,
+                     int sock_proto):
+    addrinfo = __static_getaddrinfo_pyaddr(
+        addr[0], addr[1],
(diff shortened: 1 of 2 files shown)


uvloop/handles/udp.pyx
```diff
@@ -188,3 +182,17 @@ cdef class UDPTransport(UVBaseTransport):
         # Ensure that what we buffer is immutable.
         self.buffer.append((bytes(data), addr))
         self._maybe_pause_protocol()
+
+    @functools.lru_cache()
+    def _validate_address(self, object addr):
```
1st1: Can you try to make it a standalone function (not a method)?  That way the cache will be shared across different transports.

jlaine: Absolutely, done.

1st1:
Thank you!

jlaine:
Thanks for the amazingly fast turnaround!

Issue #207, opened 2018-11-04
UDPTransport AttributeError


amelchio:
* **uvloop version**: 0.11.2
* **Python version**: 3.6
* **Platform**: Linux
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: Unknown

This error was reported [in a forum post](https://community.home-assistant.io/t/lifx-integration-breaks-hass-io-how-to-revert/76610):
```
File “uvloop/handles/udp.pyx”, line 157, in uvloop.loop.UDPTransport.sendto
AttributeError: ‘uvloop.loop.UDPTransport’ object has no attribute ‘_address’
```
It seems like the attribute should be [`address`](https://github.com/MagicStack/uvloop/blob/v0.11.2/uvloop/handles/udp.pyx#L18), not [`_address`](https://github.com/MagicStack/uvloop/blob/v0.11.2/uvloop/handles/udp.pyx#L157).

1st1:
Thanks for submitting this. Care to make a PR?

amelchio:
Not really. I am not set up for uvloop development and I don't even know how to reproduce the problem.

jlaine:
This looks like a real bug, still present in the current code:

https://github.com/MagicStack/uvloop/blob/d5ad2b86fff1bf08a391aa344561e69c98be3b66/uvloop/handles/udp.pyx#L168

PR #216, opened 2019-01-09
Don't use non-existent UDPTransport._address attribute (fixes: #207)

jlaine:
If the sendto() method is called after closing the transport, it bombs
with an AttributeError as there is no `._address` attribute. Replacing
this by `.address` is not a solution either, as we need to exit in any
case otherwise the call to sendto() bombs again due to `.sock` having
been set to None.

### DIFF
--- a/tests/test_udp.py
+++ b/tests/test_udp.py
@@ -201,6 +201,19 @@ class Test_UV_UDP(_TestUDP, tb.UVTestCase):
         s_transport.close()
         self.loop.run_until_complete(asyncio.sleep(0.01, loop=self.loop))
 
+    def test_send_after_close(self):
+        coro = self.loop.create_datagram_endpoint(
+            asyncio.DatagramProtocol,
+            local_addr=('127.0.0.1', 0),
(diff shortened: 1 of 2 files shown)


1st1:
(This PR is related to https://github.com/MagicStack/uvloop/issues/207)

Yeah, I think this `_address` check is a remnant of the old UDP implementation and the PR is correct to just drop the check.

1st1:
Thank you, Jeremy!

jlaine:
You are most welcome! Are there any blockers which need to be addressed for the 0.12.0 release which I could help with?

jlaine:
Ouch, I was hoping the performance optimization would go in as well since currently uvloop's UDP send performance is worse than vanilla asyncio, meaning you have to choose between a fast event loop and fast UDP transmission..

1st1:
> Ouch, I was hoping the performance optimization would go in as well since currently uvloop's UDP send performance is worse than vanilla asyncio

This makes it a pretty strong case to include the change in 0.12 :)

Issue #217, opened 2019-01-09
Test certificate should be upgraded to >= 2048 bits


jlaine:
* **uvloop version**: 0.12.0rc2
* **Python version**: 3.7
* **Platform**: Debian/unstable

Running the test suite on Debian/unstable (OpenSSL 1.1.1) fails when loading the test certificate with:

`ssl.SSLError: [SSL: EE_KEY_TOO_SMALL] ee key too small (_ssl.c:3816)`

I recently encountered the same error with the `aioice` test suite, the fix is to regenerate a private key + cert of 2048 bits or more.

jlaine:
Since we seem to use the Python test suite's certificates we can make use of:

https://github.com/python/cpython/commit/e6dac0077996b1e1f886f036d6f2606237fa4c85

jlaine:
I have submitted PR #218 to update the test certificate / key

PR #218, opened 2019-01-09
Update test SSL certificate and key to 3072 bit (fixes: #217)

jlaine:
On Debian/unstable the test suite fails due to the SSL key which is only
1024 bits long. We again borrow the CPython test suites certificates
which were recently updated to 3072 bits.


1st1:
Thank you, Jeremy!

PR #224, opened 2019-02-04
Implement Server.start_serving() and related APIs

jlaine:
This is still work in progress, but I wanted to have some code up so we can start discussing it.

- I have moved `backlog` to a member, so that `listen` (internal API for `start_serving`) can be called without arguments.

- I noticed some argument validation (e.g.: backlog, ssl, ssl_backlog, timeouts) is performed quite deep inside the code, inside `UVStreamServer`. Shouldn't this be done closer to the API surface?

- Currently one test fails because the previous code caught an `OSError` and re-raised it with a different message. Is this a behaviour we want to preserve?

- I'm having trouble understanding the `Server` class, which seem to serve as the user-facing API as opposed to the internal `UVStreamServer` subclasses. The bit that bothers we: why does `Server` have a **list** of servers, can there really be anything other than 0 or 1?

- Who should carry the internal `is_serving` state, the `Server` or the `UVStreamServer`?

- How should we implement the future returned by `serve_forever`?

### DIFF
--- a/uvloop/_testbase.py
+++ b/uvloop/_testbase.py
@@ -67,6 +67,10 @@ class BaseTestCase(unittest.TestCase, metaclass=BaseTestCaseMeta):
     def mock_pattern(self, str):
         return MockPattern(str)
 
+    def has_start_serving(self):
+        return not (self.is_asyncio_loop() and
+                    sys.version_info[:2] in [(3, 5), (3, 6)])
+
     def is_asyncio_loop(self):
         return type(self.loop).__module__.startswith('asyncio.')
 
--- a/uvloop/includes/stdlib.pxi
+++ b/uvloop/includes/stdlib.pxi
@@ -66,6 +66,8 @@ cdef gc_disable = gc.disable
 cdef iter_chain = itertools.chain
 cdef inspect_isgenerator = inspect.isgenerator
 
+cdef int has_IPV6_V6ONLY = hasattr(socket, 'IPV6_V6ONLY')
+cdef int IPV6_V6ONLY = getattr(socket, 'IPV6_V6ONLY', -1)
 cdef int has_SO_REUSEPORT = hasattr(socket, 'SO_REUSEPORT')
 cdef int SO_REUSEPORT = getattr(socket, 'SO_REUSEPORT', 0)
 cdef int SO_BROADCAST = getattr(socket, 'SO_BROADCAST')
@@ -118,6 +120,7 @@ cdef sys_set_coroutine_wrapper = sys.set_coroutine_wrapper
 cdef sys_get_coroutine_wrapper = sys.get_coroutine_wrapper
 cdef sys_getframe = sys._getframe
 cdef sys_version_info = sys.version_info
+cdef str sys_platform = sys.platform
 
 cdef ssl_SSLContext = ssl.SSLContext
 cdef ssl_MemoryBIO = ssl.MemoryBIO
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1084,48 +1084,6 @@ cdef class Loop:
                     sys_set_coroutine_wrapper(None)
                     self._coroutine_debug_set = False
 
-    cdef _create_server(self, system.sockaddr *addr,
-                        object protocol_factory,
-                        Server server,
-                        object ssl,
-                        bint reuse_port,
-                        object backlog,
-                        object ssl_handshake_timeout,
-                        object ssl_shutdown_timeout):
-        cdef:
-            TCPServer tcp
-            int bind_flags
-
-        tcp = TCPServer.new(self, protocol_factory, server, ssl,
-                            addr.sa_family,
-                            ssl_handshake_timeout, ssl_shutdown_timeout)
-
-        if reuse_port:
-            self._sock_set_reuseport(tcp._fileno())
-
-        if addr.sa_family == uv.AF_INET6:
-            # Disable IPv4/IPv6 dual stack support (enabled by
-            # default on Linux) which makes a single socket
-            # listen on both address families.
-            bind_flags = uv.UV_TCP_IPV6ONLY
-        else:
-            bind_flags = 0
-
-        try:
-            tcp.bind(addr, bind_flags)
-            tcp.listen(backlog)
-        except OSError as err:
-            pyaddr = __convert_sockaddr_to_pyaddr(addr)
-            tcp._close()
-            raise OSError(err.errno, 'error while attempting '
-                          'to bind on address %r: %s'
-                          % (pyaddr, err.strerror.lower()))
-        except Exception:
-            tcp._close()
-            raise
-
-        return tcp
-
     def _get_backend_id(self):
         """This method is used by uvloop tests and is not part of the API."""
         return uv.uv_backend_fd(self.uvloop)
@@ -1584,10 +1542,11 @@ cdef class Loop:
                             sock=None,
                             backlog=100,
                             ssl=None,
-                            reuse_address=None,  # ignored, libuv sets it
+                            reuse_address=None,
                             reuse_port=None,
                             ssl_handshake_timeout=None,
-                            ssl_shutdown_timeout=None):
+                            ssl_shutdown_timeout=None,
+                            start_serving=True):
         """A coroutine which creates a TCP server bound to host and port.
 
         The return value is a Server object which can be used to stop
@@ -1641,7 +1600,8 @@ cdef class Loop:
                 raise ValueError(
                     'host/port and sock can not be specified at the same time')
             return await self.create_unix_server(
-                protocol_factory, sock=sock, backlog=backlog, ssl=ssl)
+                protocol_factory, sock=sock, backlog=backlog, ssl=ssl,
+                start_serving=start_serving)
 
         server = Server(self)
 
@@ -1661,6 +1621,8 @@ cdef class Loop:
                 raise ValueError(
                     'host/port and sock can not be specified at the same time')
 
+            if reuse_address is None:
+                reuse_address = os_name == 'posix' and sys_platform != 'cygwin'
             reuse_port = bool(reuse_port)
             if reuse_port and not has_SO_REUSEPORT:
                 raise ValueError(
@@ -1680,6 +1642,7 @@ cdef class Loop:
             infos = await aio_gather(*fs, loop=self)
 
             completed = False
+            sock = None
             try:
                 for info in infos:
                     addrinfo = (<AddrInfo>info).data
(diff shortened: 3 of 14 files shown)


1st1:
> This is still work in progress, but I wanted to have some code up so we can start discussing it.

Thank you!  Sorry for the late review, I was OOO for about 10 days.  I'll try to address all your questions below:

> * I have moved `backlog` to a member, so that `listen` (internal API for `start_serving`) can be called without arguments.

OK.

> * I noticed some argument validation (e.g.: backlog, ssl, ssl_backlog, timeouts) is performed quite deep inside the code, inside `UVStreamServer`. Shouldn't this be done closer to the API surface?

Sure.  If necessary we can always add an assertion at a lower level for clarity.

> * Currently one test fails because the previous code caught an `OSError` and re-raised it with a different message. Is this a behaviour we want to preserve?

Usually we try to match (exactly) `OSError`s that Python asyncio is raising, down to the message of the exception.  Is that the case in the previous code?

> * I'm having trouble understanding the `Server` class, which seem to serve as the user-facing API as opposed to the internal `UVStreamServer` subclasses.

It is a user-facing API: the object is returned from the `create_server()` coroutine.

Things that have the `UV` prefix in the `handles` directory are Python high-level wrappers around low-level libuv concepts.

> The bit that bothers we: why does `Server` have a **list** of servers, can there really be anything other than 0 or 1?

Yes, there can be many.  If you create a server on `localhost` it can create more than one server; in the most simple case, you'll have one for IPv4 and one for IPv6.

In asyncio implementation (see `CPython/Lib/asyncio/base_events.py` for the reference) the `Server` object has a list of Python sockets that the server listens on.  In uvloop, we have a list of libuv's `uv_tcp_t` handles wrapped in `UVStreamServer` objects.  We later call `uv_listen` on them to start accepting connections.

> * Who should carry the internal `is_serving` state, the `Server` or the `UVStreamServer`?

The `Server` objects, as it is a user-facing API that controls a bunch of low-level libuv handles.

> * How should we implement the future returned by `serve_forever`?

I think we should be able to copy/paste asyncio implementation from `base_events.py`.

jlaine:
> > * Currently one test fails because the previous code caught an `OSError` and re-raised it with a different message. Is this a behaviour we want to preserve?
> 
> Usually we try to match (exactly) `OSError`s that Python asyncio is raising, down to the message of the exception. Is that the case in the previous code?

AFAICT, cpython only wraps `OSError` when calling `sock.bind()`, not `listen()`:

https://github.com/python/cpython/blob/master/Lib/asyncio/base_events.py#L943

> > The bit that bothers we: why does `Server` have a **list** of servers, can there really be anything other than 0 or 1?
> 
> Yes, there can be many. If you create a server on `localhost` it can create more than one server; in the most simple case, you'll have one for IPv4 and one for IPv6.

OK sorry for the noise, I missed the `for` loop.

> > * How should we implement the future returned by `serve_forever`?
> 
> I think we should be able to copy/paste asyncio implementation from `base_events.py`.

Indeed, I did just that.

jlaine:
I had another look at "why are the OSError messages different", and it looks like the reason is:

- The call to `uv_tcp_bind` succeeds even if there is already a socket bound to the same address / port.

- The call to `uv_listen` however fails if there is already a socket bound to the same address / port.

This is very different from plain asyncio, which bombs as soon as it attempts `socket.bind()`. This API difference used not to be visible to users, as the bind / listen steps were run in immediate succession, but with the introduction of `start_serving` it becomes important. How do you suggest we proceed?

1st1:
Oh well... I guess the only sure option we have is to rewrite this part:

https://github.com/MagicStack/uvloop/blob/d8fe153e5b67de1124b828f9c9b3c5397ac68d5d/uvloop/loop.pyx#L1654-L1696

to create Python sockets and bind them in the same way asyncio does.  After we have Python sockets bound, we can wrap them in `TCPServer` objects the same way we do it here:

https://github.com/MagicStack/uvloop/blob/d8fe153e5b67de1124b828f9c9b3c5397ac68d5d/uvloop/loop.pyx#L1708-L1713

That's the only way I can think of that will preserve full compatibility with asyncio.  Performance-wise this shouldn't have any impact.

jlaine:
OK sounds like a plan!

jlaine:
@1st1 I have implemented what you suggested, but I'm now getting a warning about an unclosed resource:

```
uvloop/tests/test_tcp.py:967: ResourceWarning: unclosed <socket.socket fd=11, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('0.0.0.0', 43443)>
  None, 0, backlog=bl)
```

What do we need to do with the Python socket object once we have wrapped it in a `TCPServer`?

uvloop/includes/stdlib.pxi
```diff
@@ -63,6 +63,8 @@ cdef gc_disable = gc.disable
 cdef iter_chain = itertools.chain
 cdef inspect_isgenerator = inspect.isgenerator
 
+cdef int IPV6_V6ONLY = getattr(socket, 'IPV6_V6ONLY')
+cdef int has_SO_REUSEADDR = os.name == 'posix' and sys.platform != 'cygwin'
```
jlaine: This is directly lifted from: https://github.com/python/cpython/blob/3e028b2d40370dc986b6f3146a7ae927bc119f97/Lib/asyncio/base_events.py#L1220

uvloop/includes/stdlib.pxi
```diff
@@ -63,6 +63,8 @@ cdef gc_disable = gc.disable
 cdef iter_chain = itertools.chain
 cdef inspect_isgenerator = inspect.isgenerator
 
+cdef int IPV6_V6ONLY = getattr(socket, 'IPV6_V6ONLY')
```
jlaine: Is this the right place / way to do it?

uvloop/loop.pyx
```diff
@@ -1674,25 +1635,68 @@ cdef class Loop:
...
+                                            ssl, ssl_handshake_timeout,
+                                            ssl_shutdown_timeout)
+
+                        try:
+                            tcp._open(sock.fileno())
+                        except Exception:
+                            tcp._close()
+                            raise
+
+                        tcp._attach_fileobj(sock)
```
1st1: I think we can try calling `sock.detach()` and skip this line.  There's no point in keeping the reference to the original socket here.  Hopefully that will fix the warning you saw.

uvloop/includes/stdlib.pxi
1st1: It's not about the platform having `SO_REUSEADDR` or not, its about it behaving quite differently on Windows.  Coincidentally there's a relevant discussion about this option here: https://github.com/python/cpython/pull/11784#discussion_r254843057 if you are interested what's going on.

Regarding `has_SO_REUSEADDR` -- I'd just inline this check to `create_server()`.

uvloop/includes/stdlib.pxi
1st1: It is, but `IPV6_V6ONLY` might be missing on some platforms.  So you probably want to add two constants: `has_IPV6_ONLY = hasattr(socket, 'IPV6_V6ONLY')` and `IPV6_V6ONLY = getattr(socket, 'IPV6_V6ONLY', -1)`

1st1: Context: this whole ugly file is the result of Cython not cleaning up its PyModules properly.  That lead to circular references that prevented uvloop ResourceWarnings from showing (say about an event loop that the user forgot to properly close).  So this file is a workaround for that bug -- it references stuff from the stdlib without keeping references to Python modules.

Now this Cython shortcoming has been fixed in some recent Cython release, so we probably can start importing things properly. It's something for us to check later.

uvloop/includes/stdlib.pxi
1st1: (warning: the linked discussion is a bit... "inflammatory" 🤨)

uvloop/loop.pyx
jlaine: OK I'll do a `sock.detach()` followed by `sock = None`.

Regarding the warning I saw, it was something else: if any calls raise an exception we need to ensure the socket gets closed. That is why I keep track of `sock` and close it in the `finally`, otherwise we need try/catch every function call and do sock.close() at every possible failure.

uvloop/includes/stdlib.pxi
jlaine: I'm not sure how to inline the check, I had done so initially but the code bombed, even if I `import os` and `import sys` I don't seem to be able to use `os.name` or `sys.platform`.. some Cython pecularity I don't know?

uvloop/includes/stdlib.pxi
jlaine: OK

uvloop/includes/stdlib.pxi
1st1: just add `os_name = os.name` and `sys_platform = sys.platform` to `uvloop/includes/stdlib.pxi`.  We'll later than be able to use search/replace to remove that file.

uvloop/includes/stdlib.pxi
1st1: No, it's the same. It's just an inconsistency that we'll want to fix at some point.  Initially my idea was to use Python constants for Python objects/APIs and `uv.*` constants for libuv, but I guess it's actually more confusing this way.

uvloop/_testbase.py
```diff
@@ -67,6 +67,9 @@ def new_loop(self):
     def mock_pattern(self, str):
         return MockPattern(str)
 
+    def has_start_serving(self):
+        return not (self.is_asyncio_loop() and sys.version_info[:2] in [(3, 5), (3, 6)])
```
1st1: Everything looks great. One last nit: please make sure all lines are shorter than 80 chars.

jlaine: Sure thing. I wasn't too sure as the limit doesn't seem to be 100% enforced. Maybe we should get CI to enforce this ?

1st1: Sure: https://github.com/MagicStack/uvloop/pull/229 :)

1st1: Speaking of which, let's land https://github.com/MagicStack/uvloop/pull/229 first and then rebase this one.

jlaine:
Hi Yury, what's the commit policy for uvloop, specifically who is expected to actually do the merge to master? The reviewer, author, .. ?

1st1:
You can merge it yourself. Please rebase it on top of recent master before merging though (has now flake8 integration)

jlaine:
OK I've rebased on top of master and fixed 2 flake8 errors. If CI completes successfully I'll merge it.

denismakogon:
When a new release with these changes will be published?

1st1:
Need to implement sendfile support first.

denismakogon:
Any estimates for that?

1st1:
Unfortunately I don't have time to tackle that myself in the next couple of weeks. Hopefully we'll get it in March.

1st1:
Thanks so much, @jlaine!

PR #225, opened 2019-02-11
Fix ref cycles in callback handles

1st1:


1st1:
@fantix please take a look

PR #229, opened 2019-02-15
Enforce consistent code style with flake8

1st1:


jlaine:
I wonder where this is coming from :)

1st1:
Actually we do run flake8 in our other projects as part of CI, I just didn't know we can use flake8 on Cython files.  A nice TIL ;)

1st1:
Thanks for nudging me to the right direction though :)

PR #230, opened 2019-02-15
Use travis homebrew add-on to speed up OS X builds

jlaine:
The OS X builds take a long time upgrading homebrew manually, which can
be avoided by using the "homebrew" add-on.

### DIFF
--- a/.ci/travis-install.sh
+++ b/.ci/travis-install.sh
@@ -3,8 +3,6 @@
 set -e -x
 
 if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
-    brew update >/dev/null
-    brew upgrade pyenv
     eval "$(pyenv init -)"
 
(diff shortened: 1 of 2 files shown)


jlaine:
Well that failed miserably..

1st1:
Should this one be closed? Or there's something else we can try?

elprans:
Try adding `update: true` to the `homebrew` hash

jlaine:
Adding "update: true" kind of defeats the point of what I'm trying to do here, which is to cut out the lengthy update process!

elprans:
We are deliberately choosing an older version to ensure compatibility with a couple of prior macOS releases.  If there's a good way of achieving that on newer releases, I'm all for that.

1st1:
Looks like CI isn't any faster with this PR, right?

jlaine:
Nope, because we are again updating homebrew.

Issue #221, opened 2019-01-28
Python 3.7 "TCP echo server using streams" sample code does not work with uvloop.


dstruck:
* **uvloop version**: 0.12.0
* **Python version**: 3.7.1
* **Platform**: Linux
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: not tested

Running the following code (source: https://docs.python.org/3/library/asyncio-stream.html#tcp-echo-server-using-streams) with uvloop:

```python
import asyncio
import uvloop
import sys

async def handle_echo(reader, writer):
  data = await reader.read(100)
  message = data.decode()
  addr = writer.get_extra_info('peername')

  print(f"Received {message!r} from {addr!r}")

  print(f"Send: {message!r}")
  writer.write(data)
  await writer.drain()

  print("Close the connection")
  writer.close()

async def main():
  server = await asyncio.start_server(handle_echo, '127.0.0.1', 8888)

  addr = server.sockets[0].getsockname()
  print(f'Serving on {addr}')

  async with server:
    await server.serve_forever()

uvloop.install()

print("python version:", sys.version)
print("uvloop version:", uvloop.__version__)

asyncio.run(main())
```

Produces the following error message:

```bash
python uvlopp_issue.py 
python version: 3.7.1 (default, Dec 14 2018, 19:28:38) 
[GCC 7.3.0]
uvloop version: 0.12.0
Serving on ('127.0.0.1', 8888)
Traceback (most recent call last):
  File "uvlopp_issue.py", line 36, in <module>
    asyncio.run(main())
  File "/home/test/miniconda3/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "uvloop/loop.pyx", line 1451, in uvloop.loop.Loop.run_until_complete
  File "uvlopp_issue.py", line 27, in main
    async with server:
AttributeError: __aexit__
```

1st1:
Thank you for reporting.  This is basically a duplicate of https://github.com/MagicStack/uvloop/issues/211.

jlaine:
This was fixed by PR #224

ahesford:
I don't believe PR #224 fixes this issue. This issue is caused because the stock asyncio.base_events.Server in Python 3.7 was made into an asynchronous context manager. The entry method is simple:

    async def __aenter__(self):
        return self

while the exit method closes the server:

    async def __aexit__(self, *exc):
        self.close()
        await self.wait_closed()

[...]

PR #234, opened 2019-03-16
Make Server an asynchronous context manager (fixes: #221)

jlaine:
The Server class was missing `__aenter__` and `__aexit__` magic methods to
allow usage of the form:

```
  async with server:
    await server.serve_forever()
```


1st1:
Thanks!

PR #236, opened 2019-03-17
Round delay in call_later()

### LINKED ISSUE #233: Incorrect time calculation in loop.call_at()
* **uvloop version**: 0.11.2
* **Python version**: 3.6.7
* **Platform**: ubuntu 18.10

There is a conceptual bug in the calculation of the time in call_at() here:
https://github.com/MagicStack/uvloop/blob/master/uvloop/loop.pyx#L1270
If an application calculates a time in the future to call a function with call_at(), it should be able to expect
the value of loop.time() when called inside that function to be equal or higher (later) than the timestamp that was specified. Due to timer resolution and the fact that call_at() internally does "when - self.time()" causes that this assumption is not guaranteed to be true.
The following script demonstrates this:

```python
#!/usr/bin/env python3

import sys
import asyncio
import uvloop

def mytimer():
        global ti
        global loop
        ti0 = loop.time()
        print(ti, ti0)
        if ti > ti0:
                print("ERROR!!!")
                sys.exit(1)
        ti = ti0 + 0.1
        loop.call_at(ti, mytimer)


if __name__ == "__main__":
        global ti
        global loop
        ti = loop.time()
        loop.call_at(ti + 1.0, mytimer)
        loop.run_forever()
```
Eventually this script will stop after printing "ERROR!!!". This means that the timer handler mytimer() was
called a bit too early.

fantix:
- [x] Write unit test.
- [x] Make the test pass.

* Fixes #233.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -171,6 +171,19 @@ class _TestBase:
         self.loop.run_forever()
         self.assertEqual(calls, ['a'])
 
+    def test_call_later_rounding(self):
+        # Refs #233, call_later() and call_at() shouldn't call cb early
+
+        def cb():
+            self.loop.stop()
+
+        for i in range(8):
+            self.loop.call_later(0.06 + 0.01, cb)  # 0.06999999999999999
+            started = int(round(self.loop.time() * 1000))
+            self.loop.run_forever()
+            finished = int(round(self.loop.time() * 1000))
+            self.assertGreaterEqual(finished - started, 70)
+
     def test_call_at(self):
         if os.environ.get('TRAVIS_OS_NAME'):
             # Time seems to be really unpredictable on Travis.
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1254,7 +1254,7 @@ cdef class Loop:
             # infinity for a Python application.
             delay = 3600 * 24 * 365 * 100
 
-        when = <uint64_t>(delay * 1000)
+        when = <uint64_t>round(delay * 1000)
         if not args:
             args = None
         if when == 0:


jlaine:
If we want to respect the guarantee that the CB is called "no earlier than X", isn't ceil() more appropriate than round()?

fantix:
@jlaine replied in https://github.com/MagicStack/uvloop/issues/233#issuecomment-473678282

jlaine:
Ok sorry for the duplicate message

1st1:
So is the new test stable?

fantix:
Yeah I think so

1st1:
Alright, merged. Thank you! :)

1st1:
@fantix got this CI failure this morning:

Screen Shot 2019-03-20 at 10 55 45 AM

Could you please take another look at the test?  Is it possible to make it any more reliable?  If not, I suggest we relax it a bit or drop it.

1st1:
(or maybe we should go with @jlaine suggestion to use `ceil()`)

fantix:
@1st1 `ceil()` may not fix this one - there's arbitrary time spent before the clock starts ticking. I think an absolute safe number would be `69`, but it's probably not much meaningful anymore. For what is worth, maybe `69` for now?

1st1:
Let's try it... Would the old code (without `round()`) fail on that?

fantix:
@1st1 the old code may probably succeed with this test at `69` I think

1st1:
Alright, I pushed https://github.com/MagicStack/uvloop/commit/e3b00b81024fd69de9b4d51a8ae49ffe033bc0cb

fantix:
Oh, thanks for that! 🙇

PR #238, opened 2019-03-17
Start using high-performance uv_udp_t handle again

1st1:
This reverts the UDP implementation to use the `uv_udp_t` handle.  This should be much faster than using `uv_poll_t` handles.

The only caveat is that this requires libuv to support `AF_UNIX` sockets in `uv_udp_send()`.  We'll need to create a PR for that.  Meanwhile, the following patch can be used to test this PR:

```diff
diff --git a/src/unix/udp.c b/src/unix/udp.c
index 15fa5937..b6d3d0f1 100644
--- a/src/unix/udp.c
+++ b/src/unix/udp.c
@@ -30,6 +30,7 @@
 #if defined(__MVS__)
 #include <xti.h>
 #endif
+#include <sys/un.h>

 #if defined(IPV6_JOIN_GROUP) && !defined(IPV6_ADD_MEMBERSHIP)
 # define IPV6_ADD_MEMBERSHIP IPV6_JOIN_GROUP
@@ -232,8 +233,16 @@ static void uv__udp_sendmsg(uv_udp_t* handle) {
       h.msg_namelen = 0;
     } else {
       h.msg_name = &req->addr;
-      h.msg_namelen = req->addr.ss_family == AF_INET6 ?
-        sizeof(struct sockaddr_in6) : sizeof(struct sockaddr_in);
+      if (req->addr.ss_family == AF_INET6) {
+        h.msg_namelen = sizeof(struct sockaddr_in6);
+      } else if (req->addr.ss_family == AF_INET) {
+        h.msg_namelen = sizeof(struct sockaddr_in);
+      } else if (req->addr.ss_family == AF_UNIX) {
+        h.msg_namelen = sizeof(struct sockaddr_un);
+      } else {
+        assert(0 && "unsupported address family");
+        abort();
+      }
     }
     h.msg_iov = (struct iovec*) req->bufs;
     h.msg_iovlen = req->nbufs;
diff --git a/src/uv-common.c b/src/uv-common.c
index 94dd59fb..55ea0d0e 100644
--- a/src/uv-common.c
+++ b/src/uv-common.c
@@ -34,6 +34,7 @@
 # include <malloc.h> /* malloc */
 #else
 # include <net/if.h> /* if_nametoindex */
+# include <sys/un.h> /* AF_UNIX, sockaddr_un */
 #endif


@@ -376,6 +377,10 @@ int uv__udp_check_before_send(uv_udp_t* handle, const struct sockaddr* addr) {
       addrlen = sizeof(struct sockaddr_in);
     else if (addr->sa_family == AF_INET6)
       addrlen = sizeof(struct sockaddr_in6);
+#ifdef AF_UNIX
+    else if (addr->sa_family == AF_UNIX)
+      addrlen = sizeof(struct sockaddr_un);
+#endif
     else
       return UV_EINVAL;
   } else {
```


1st1:
@jlaine Jeremy, could you please try this PR?  Is it any faster than what we have now?

uvloop/handles/udp.pyx
```diff
@@ -1,199 +1,373 @@
...
+            raise convert_error(err)
+
+        if family in (uv.AF_INET, uv.AF_INET6):
+            self._family = family
+
+        self._finish_init()
+
+    cdef _connect(self, system.sockaddr* addr, size_t addr_len):
+        cdef int err
+        uv.uv_udp_connect(<uv.uv_udp_t*>self._handle, addr)
```
jlaine: It looks like `err` should be assigned?

1st1: Yeah, good catch! I've pushed a commit with a fix.

jlaine:
I was able to build your branch but I need some time to come up with a proper benchmark. My usual benchmark (DataChannel / SCTP / DTLS / ICE) won't cut it: the sockets aren't "connected" as remote candidates can come in at any time.

1st1:
> the sockets aren't "connected" as remote candidates can come in at any time.

Maybe the usual benchmark would be enough? I don't think we need to benchmark "connected" UDP sockets specifically, we're just interested to see if UDP with this PR is generally faster than in uvloop 0.12.

jlaine:
I'm not sure what has changed since last time I ran my tests, but I'm seing a behaviour I had not previously identified (both with 0.12 and with this PR):

```python
import asyncio
import socket
import time

import uvloop


class DummyProtocol(asyncio.DatagramProtocol):
    pass


async def run(loop):
    addr = ('127.0.0.1', 1234)
    data = b'M' * 1500
    transport, protocol = await loop.create_datagram_endpoint(
        lambda: DummyProtocol(),
        family=socket.AF_INET)

    for i in range(1000000):
        transport.sendto(data, addr)
    print('after sendto', time.process_time() - start)

start = time.process_time()
loop.run_until_complete(run(loop))
print('after run', time.process_time() - start)```

Both 0.12 and this PR give me almost identical results which look like:

```
after sendto 1.394025113
after run 4.101789539
```

Vanilla asyncio looks like:

```
after sendto 3.545451784
after run 3.54558408
```

The behaviour I hadn't spotted: more time seems to be spent exiting the `run()` method than actually executing it. Is this buffers being flushed?

**Comment by 1st1** [comment]
> Is this buffers being flushed?

Maybe... Can you add `transport.abort()` (closes the transport instantly) or `transport.close()` (allows the buffers to flush) and see what happens?

**Comment by 1st1** [comment]
>> Is this buffers being flushed?
> Maybe... Can you add transport.abort() (closes the transport instantly) or transport.close() (allows the buffers to flush) and see what happens?

**Comment by 1st1** [comment]
>> Is this buffers being flushed?
>
> Maybe... Can you add transport.abort() (closes the transport instantly) or transport.close() (allows the buffers to flush) and see what happens?

Alright, I think I've figured it out.

1. asyncio tries to do `sock.send()` if the send buffer is empty. It's an optimization that we have for our TCP layer, but not for the UDP.

2. Once I started to call `uv_udp_try_send()` the weird behaviour you spotted disappeared, and uvloop started to behave just like vanilla asyncio.

After some debugging and profiling I think that the delay is caused by the system call itself.  In other words, 1000000 system calls can be quite expensive.  We'll need a more elaborate benchmark that actually sends and reads data.

I also optimized how Python socket addresses are converted to low-level `sockaddr*` structs.  Now we have an LRU cache for that, which seems to add a noticeable speed boost.

**Comment by jlaine** [comment]
`__convert_pyaddr_to_sockaddr` : very nice, now all such calls benefit from the caching logic, and I see we gained a cythonized LRU cache in the process. I'm wondering if cpython's `getsockaddrarg` would also benefit from similar logic?

I still don't quite understand how the cost system calls shows up at the *end* of the `run` loop?

Using the "dumb" benchmark uvloop is now consistently ~ 20% faster than vanilla asyncio so we're definitely making progress. I'll now run some tests closer to real-world usage using:

https://github.com/aiortc/aiortc/tree/master/examples/datachannel-filexfer

(NOTE: it benchmarks much more than just raw UDP throughput as there is also DTLS + a pure-python SCTP stack involved)

**Comment by jlaine** [comment]
As promised here are some results using `aiortc` to send a file over SCTP / DTLS / UDP.

Preliminary setup
------------------------

Clone aiortc:

```
git clone https://github.com/aiortc/aiortc
cd aiortc
pip install -e .
```

Switch CPU scaling governor to performance for more stable results:

```
echo performance > /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
echo performance > /sys/devices/system/cpu/cpu1/cpufreq/scaling_governor
echo performance > /sys/devices/system/cpu/cpu2/cpufreq/scaling_governor
echo performance > /sys/devices/system/cpu/cpu3/cpufreq/scaling_governor
```

Running the tests
------------------------

Create 256MB of random data:

```dd if=/dev/urandom of=/tmp/test.bin bs=1M count=256```

Launch the sender, using a UNIX socket for signaling:

```python examples/datachannel-filexfer/filexfer.py -s unix-socket send /tmp/test.bin```

Launch the receiver:

```python examples/datachannel-filexfer/filexfer.py -s unix-socket receive /dev/null```

Results
----------

Vanilla asyncio: 120Mbps
uvloop 0.12.1: 140Mbps
uvloop (this PR): 185Mbps

I'd say that's a pretty neat performance boost!

1st1:
> I still don't quite understand how the cost system calls shows up at the end of the run loop?

Because the previous version of this PR and the current master branch only *queue* system calls in `udp.sendto()`.  The latest iteration of this PR and asyncio actually try to perform system calls right in `udp.sendto()` if there're no pending buffers to send.

1st1:
>Vanilla asyncio: 120Mbps
uvloop 0.12.1: 140Mbps
uvloop (this PR): 185Mbps
>
>I'd say that's a pretty neat performance boost!

Great news! :)  A microbenchmark would probably show even greater performance improvement.

1st1:
> A microbenchmark would probably show even greater performance improvement.

I mean a microbenchmark that actually does receive and send data, similar to TCP echo server bench in `examples/bench/echoserver.py`.  The benchmark must engage the rest of uvloop/asyncio transport/protocol code.  Benchmarking how efficiently they do system calls doesn't tell us much.

1st1:
Good news, our PR to libuv  https://github.com/libuv/libuv/pull/2220 has been merged. Waiting for the release now.

jlaine:
Awesome, fingers crossed :)

In case you're interested I've started a pure-python QUIC implementation. Still early days but it will eventually also provide a good usecase for fast asyncio / UDP:

https://github.com/aiortc/aioquic

1st1:
That's a great project, Jeremy. We need that for Python, keep me updated on it.

jlaine:
.. and libuv 1.28 is out with your patch!

1st1:
Merged to `master`. I'm thinking about making an "RC" release in hopes that someone will actually test it and find bugs in UDP.

Issue #251, opened 2019-06-07
sys.set_coroutine_wrapper will be removed in Python 3.8


encukou:
* **uvloop version**: master (cbbac15f653f08308db025c74d6ce1f474a895f4)
* **Python version**: 3.8.0b1
* **Cython version**: 0.29.10 (updated for Python 3.8 support)
* **Platform**: Fedora
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: yes

The function "sys.set_coroutine_wrapper", which was *provisionally* added to Python 3.5 and [deprecated in Python 3.7](https://bugs.python.org/issue32591), will be [removed in Python 3.8](https://docs.python.org/3.8/whatsnew/3.8.html#api-and-feature-removals). If uvloop needs it, now is the last chance to make the case for it.

When compiled with Python 3.8.0b1, uvloop fails to import:

```
Python 3.8.0b1 (default, Jun  5 2019, 14:47:51) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import uvloop
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pviktori/dev/uvloop/uvloop/__init__.py", line 7, in <module>
    from .loop import Loop as __BaseLoop  # NOQA
  File "uvloop/includes/stdlib.pxi", line 119, in init uvloop.loop
    cdef sys_set_coroutine_wrapper = sys.set_coroutine_wrapper
AttributeError: module 'sys' has no attribute 'set_coroutine_wrapper'
```

1st1:
No, we don't need it. We'll have to fix the compatibility with 3.8 at some point though.

hroncok:
AFAIK It's only used conditionally in here:

https://github.com/MagicStack/uvloop/blob/46c5e9ebe45c7d5aa8777ad1add6656e349528ab/uvloop/loop.pyx#L1055-L1087

hroncok:
My Cython knowledge is limited here. I wanted to do something like:

```cython
from cpython.version cimport PY_VERSION_HEX
...

IF PY_VERSION_HEX > 0x03080000:
    cdef sys_set_coroutine_wrapper = sys.set_coroutine_wrapper
    cdef sys_get_coroutine_wrapper = sys.get_coroutine_wrapper
```

[...]

PR #253, opened 2019-06-10
sys.set_coroutine_wrapper is removed in Python 3.8

hroncok:
The functions `sys.set_coroutine_wrapper` and `sys.get_coroutine_wrapper`,
which were provisionally added to Python 3.5 and deprecated in Python 3.7,
are removed in Python 3.8.

uvloop doesn't use them since Python 3.7, but they remained defined
in stdlib.pxi, leading to AttributeErrors on Python 3.8+.

This makes the definition fallback to None if such functions don't exist.

Fixes https://github.com/MagicStack/uvloop/issues/251


hroncok:
Rebased.

1st1:
Thank you!

PR #263, opened 2019-08-14
add app state check

### LINKED ISSUE #246: SSLProtocol violates app Protocol state machine
* **uvloop version**: 0.12.2
* **Python version**: 3.5.4
* **Platform**: Mac OS X

Some instances of the `Protocol` class assume `connection_lost()` can only be called if `connection_made()` was called at some point. This does not appear to be explicitly guaranteed in the Python documentation, but the following comment strongly suggests that intent:
https://github.com/python/cpython/blob/master/Lib/asyncio/protocols.py#L84

An error raised during the handshake phase triggers a transition back to `UNWRAPPED`:
https://github.com/MagicStack/uvloop/blob/master/uvloop/sslproto.pyx#L504
Following this code path further, we see that it will force close the transport, which triggers a call back to `SSLProtocol.connection_lost()`. This is then forwarded onto the app's `Protocol` instance, even though `connection_made()` was never called.

Would it be appropriate to change the following guard to also exclude the `UNWRAPPED` state?
https://github.com/MagicStack/uvloop/blob/master/uvloop/sslproto.pyx#L337
Alternatively a boolean value could be used to track whether or not `connection_made()` has ever been called.

fantix:
* fixes #246 

`app_state` check is not added at `data_received()` and friends because I think it's too much penalty to raise errors and break connections, even though it should never happen. But please feel free to vote oppositely.

AIO State

- [x] write tests

### DIFF
--- a/uvloop/sslproto.pxd
+++ b/uvloop/sslproto.pxd
@@ -6,6 +6,20 @@ cdef enum SSLProtocolState:
     SHUTDOWN = 4
 
 
+cdef enum AppProtocolState:
+    # This tracks the state of app protocol (https://git.io/fj59P):
+    #     start -> CM [-> DR*] [-> ER?] -> CL -> end
+    # * CM: connection_made()
+    # * DR: data_received()
+    # * ER: eof_received()
+    # * CL: connection_lost()
+
+    START = 0
+    AFTER_CM = 1
+    AFTER_ER = 2
+    END = 3
+
+
 cdef class _SSLProtocolTransport:
     cdef:
         object _loop
@@ -30,7 +44,6 @@ cdef class SSLProtocol:
         bint _app_transport_created
 
         object _transport
-        bint _call_connection_made
         object _ssl_handshake_timeout
         object _ssl_shutdown_timeout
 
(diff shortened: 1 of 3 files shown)


uvloop/sslproto.pxd
```diff
@@ -6,6 +6,20 @@ cdef enum SSLProtocolState:
...
+    # This tracks the state of app protocol (https://git.io/fj59P):
+    #     start -> CM [-> DR*] [-> ER?] -> CL -> end
+    # * CM: connection_made()
+    # * DR: data_received()
+    # * ER: eof_received()
+    # * CL: connection_lost()
+
+    START = 0
+    AFTER_CM = 1
+    AFTER_ER = 2
```
1st1: Can we rename these to `STATE_INIT`, `STATE_CON_MADE`, `STATE_EOF`, `STATE_CON_LOST`? Also the comment isn't synced with enum members.

fantix: Makes sense to me! Updated with a state diagram, and fixed the comment too.

1st1:
Looks good, thank you!

PR #270, opened 2019-09-26
Restore wakeup_fd if it was updated in _recv_signals_start

vladima:
Common behavior for event loops bundled with CPython: they call `signal.set_wakeup_fd` with event loop's pipe descriptor in `add_signal_handler` so if no signal handlers were added - wakeup fd is not modified. UVLoop works differently and eagerly sets wakeup fd inside `_recv_signals_start` called from `run` to make sure that default signal handlers will be invoked. In general this difference does not cause any problems however for some corner cases it becomes an issue. 

As an example: during migration of the large codebase to asyncio sometimes it is hard to convert the entire callstack from sync to async (i.e. because code in question is invoked from some library bits that are not yet async friendly). In order to do this we create nested event loops to start async calls from sync functions and wait for  their completion. That works reasonably well, however in case if top level event loop sets up signal handler (and thus sets its self pipe as `wakeup_fd`) and then somewhere down the stack nested uvloop is created (with no signal handlers attached) - its self pipe overwrites the `wakeup_fd` so top level event loop will never receive the signal. For in-house event loops this issue does not occur because they don't eagerly set `wakeup_fd`.

This PR brings behavior of UVLoop a bit closer to asyncio event loops by restoring `wakeup_fd` if it was changed eagerly.

### DIFF
--- a/tests/test_signals.py
+++ b/tests/test_signals.py
@@ -275,6 +275,46 @@ finally:
         with self.assertRaisesRegex(TypeError, 'coroutines cannot be used'):
             self.loop.add_signal_handler(signal.SIGHUP, coro)
 
+    def test_wakeup_fd_unchanged(self):
+        async def runner():
+            PROG = R"""\
+import uvloop
+import signal
+import asyncio
+
+
+def get_wakeup_fd():
+    fd = signal.set_wakeup_fd(-1)
+    signal.set_wakeup_fd(fd)
+    return fd
+
+async def f(): pass
+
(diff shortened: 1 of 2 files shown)


1st1:
Looks alright. I'll wait until the CI is green though.

1st1:
Could you please rebase on top of latest master?

1st1:
Thank you! Sorry it took a bit longer than usual. Will be part of 0.14.

PR #280, opened 2019-10-21
fix test_renegotiation

fantix:
- [ ] fix test in Travis

### DIFF
--- a/tests/test_tcp.py
+++ b/tests/test_tcp.py
@@ -2246,7 +2246,7 @@ class _TestSSL(tb.SSLTestCase):
         A_DATA = b'A' * 1024 * 1024
         B_DATA = b'B' * 1024 * 1024
 
-        sslctx = openssl_ssl.Context(openssl_ssl.SSLv23_METHOD)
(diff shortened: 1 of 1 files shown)


1st1:
Beautiful, thanks, Fantix! Any other places to fix tests like that in uvloop?

fantix:
Searched the code base, found a few `PROTOCOL_SSLv23` - the most widely-used one is in `_create_server_ssl_context`. I think we could replace it with `PROTOCOL_TLS` when [Python 3.5](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_SSLv23) support is dropped (unlikely?), or do you prefer supporting both?

1st1:
> or do you prefer supporting both?

We probably need to support 3.5 for this upcoming release and then we can sunset it.

1st1:
>  I think we could replace it with PROTOCOL_TLS when Python 3.5 

Can we do that now and keep using `PROTOCOL_SSLv23` for 3.5? Is this a good idea?

fantix:
Yeah sure, just a sec

Issue #277, opened 2019-10-16
uvloop build failed on Python 3.8


Snawoot:
* **uvloop version**: 0.13.0
* **Python version**: 3.8
* **Platform**: Linux (Ubuntu Xenial, Fedora 30)
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: No, can't even build uvloop

Hello,

Two days ago new stable Python version 3.8.0 has been released. Today I updated test matrix for one of my projects and noticed uvloop fails to install: https://travis-ci.org/Snawoot/postfix-mta-sts-resolver/jobs/598694743

Steps to reproduce:

```sh
pip3.8 install uvloop
```

sonirico:
Same here: https://travis-ci.com/sonirico/wpoke/jobs/246022440 
This is a big stopper in regards to make libraries take advantage of the new features introduced in py38. Fast feedback on this is much appreciated...

Jamim:
It looks like this issue duplicates #266.

Snawoot:
@Jamim Yes, it is, but this issue also escalates problem. Now uvloop is not usable on current stable python.

AHLinJie:
uvloop/loop.c:334:75: note: in definition of macro ‘__Pyx_PyCode_New’
               PyCode_New(a, 0, k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos)
                                                                               ^~~~~
    In file included from /usr/local/include/python3.8/compile.h:5:0,
                     from /usr/local/include/python3.8/Python.h:138,
                     from uvloop/loop.c:20:
    /usr/local/include/python3.8/code.h:122:28: note: expected ‘PyObject * {aka struct _object *}’ but argument is of type ‘int’
     PyAPI_FUNC(PyCodeObject *) PyCode_New(
                                ^~~~~~~~~~
    uvloop/loop.c:334:11: error: too many arguments to function ‘PyCode_New’
               PyCode_New(a, 0, k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos)
               ^
    uvloop/loop.c:166863:15: note: in expansion of macro ‘__Pyx_PyCode_New’
         py_code = __Pyx_PyCode_New(
                   ^
    In file included from /usr/local/include/python3.8/compile.h:5:0,
                     from /usr/local/include/python3.8/Python.h:138,
                     from uvloop/loop.c:20:
    /usr/local/include/python3.8/code.h:122:28: note: declared here
[...]

Issue #266, opened 2019-08-31
issue building in Python 3.8.0b4+ (a recent development version)


wumpus:
* **uvloop version**: 0.13.0
* **Python version**: 3.8.0b4+
* **Platform**: travis-ci, Ubuntu xenial 16.04.6
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: dunno

Python 3.8 development has reached the point where it has breaking changes.

PyCode_New() seems to have changed function signature.

```
uvloop/loop.c:334:11: error: too many arguments to function ‘PyCode_New’
             PyCode_New(a, 0, k, l, s, f, code, c, n, v, fv, cell, fn, name, fline, lnos)
             ^
  uvloop/loop.c:154727:34: note: in expansion of macro ‘__Pyx_PyCode_New’
     __pyx_codeobj__13 = (PyObject*)__Pyx_PyCode_New(1, 0, 3, 0, CO_OPTIMIZED|CO_NEWLOCALS, __pyx_empty_bytes, __pyx_empty_tuple, __pyx_empty_tuple, __pyx_tuple__12, __pyx_empty_tuple, __pyx_empty_tuple, __pyx_kp_s_uvloop_loop_pyx, __pyx_n_s_callback, 850, __pyx_empty_bytes); if (unlikely(!__pyx_codeobj__13)) __PYX_ERR(2, 850, __pyx_L1_error)
                                    ^
  In file included from /opt/python/3.8-dev/include/python3.8/compile.h:5:0,
                   from /opt/python/3.8-dev/include/python3.8/Python.h:138,
                   from uvloop/loop.c:20:
  /opt/python/3.8-dev/include/python3.8/code.h:122:28: note: declared here
   PyAPI_FUNC(PyCodeObject *) PyCode_New(
                              ^
```

In the past with other projects, the issue was that CPython output was checked-in and needed to be re-generated.

For an example from the Python 3.7 era: https://github.com/yaml/pyyaml/issues/69

And an example with Python 3.8, this time the CPython people changed their minds back: https://github.com/yaml/pyyaml/issues/322

1st1:
We probably need to use a newer Cython.  We'll do so soon.

RobertoPrevato:
I just compiled uvloop for Python 3.8, I didn't have issues. Instead, I got this error at runtime:

```bash
[...]

PR #284, opened 2019-10-24
Fix uvloop to run on 3.8

1st1:
Partially addresses: #277 and #266.


uvloop/_testbase.py
```diff
@@ -415,7 +415,9 @@ def __init__(self, test, sock, prog, timeout):
     def run(self):
         try:
             self._prog(TestSocketWrapper(self._sock))
-        except Exception as ex:
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except BaseException as ex:
```
elprans: Maybe do `except Exception, asyncio.CancelledError` instead.  Catching `BaseException` always makes me uneasy.

1st1: Why? We handle `KeyboardInterrupt` and `SystemExit`. Besides, that's how we do it in asyncio now, so there's no option, really :(

.travis.yml
```diff
@@ -66,6 +71,13 @@ matrix:
           python: "3.7"
           env: BUILD=tests
 
+        - os: linux
+          dist: xenial
+          sudo: required
```
Jamim: Hello @1st1,
`sudo` option [is deprecated](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).
```suggestion
```

Jamim: @1st1 I see you removed `sudo` options.
Just want to note that one of them [is still here](https://github.com/MagicStack/uvloop/blob/0c2cb54dbbf7a7661b912e76e2ae7d35e68e1922/.travis.yml#L67).

1st1: Thanks!

PR #306, opened 2019-12-18
Restore context on listen in UVStreamServer. Fix #305

### LINKED ISSUE #305: contextvars.Context not activated for connection_made()
* **uvloop version**: 0.14.0, reproducible on master (35f82506d19ef2602053304b36b93d06111d25af)
* **Python version**: 3.8
* **Platform**: Linux
* **Does uvloop behave differently from vanilla asyncio? How?**: Yes:

```python
import asyncio
import contextvars

import aiohttp
import aiohttp.web
import uvloop


var = contextvars.ContextVar('uvloop is awesome', default=None)


async def handle(request):
    assert var.get() is not None
    return aiohttp.web.Response(text='OK!')


async def client():
    await asyncio.sleep(1)

    async with aiohttp.ClientSession() as session:
        async with session.get('http://localhost:12345/') as response:
            text = await response.text()
            assert text == 'OK!'


async def setup(app):
    var.set('It truly is!')


async def create_app():
    app = aiohttp.web.Application()
    app.on_startup.append(setup)
    app.add_routes([aiohttp.web.get('/', handle)])

    asyncio.create_task(client())

    return app

if __name__ == '__main__':
    aiohttp.web.run_app(create_app(), port=12345)
```

Code works as expected on asyncio loop (`var` has value in handler) but fails with uvloop. I tracked difference up to `connection_made()`:
```
  File "uvloop/cbhandles.pyx", line 73, in uvloop.loop.Handle._run
  File "uvloop/handles/tcp.pyx", line 153, in uvloop.loop.TCPTransport._call_connection_made
  File "uvloop/handles/basetransport.pyx", line 134, in uvloop.loop.UVBaseTransport._call_connection_made
  File "uvloop/handles/basetransport.pyx", line 131, in uvloop.loop.UVBaseTransport._call_connection_made
  File "aiohttp/web_protocol.py", line 223, in connection_made
```

There is similar difference in `UVStreamHandler._on_listen`:
https://github.com/MagicStack/uvloop/blob/35f82506d19ef2602053304b36b93d06111d25af/uvloop/handles/streamserver.pyx#L67

`protocol_handler()` is being called without context activation, while with asyncio it receives context. I'm not sure if this can be the root cause.

versusvoid:


tests/test_context.py
```diff
@@ -139,6 +140,33 @@ class TrackMe:
...
+            s = socket.socket(socket.AF_INET)
+            with s:
+                s.setblocking(False)
+                await self.loop.sock_connect(s, ('127.0.0.1', port))
+
+            srv.close()
+            await srv.wait_closed()
+
+            # Let all transports shutdown.
+            await asyncio.sleep(0.1)
```
fantix: Using `sleep` is generally not recommended for new tests in uvloop - this test is actually unstable, I could get a false OK if the `srv.close()` got called before the connection is fully established.

A proper fix is to use a `Future`, set in `factory()` and awaited in `test()` - we could even pass over the assertion error in this manner. With this fix, the `ResourceWarning` is also gone.

versusvoid: Ok, fixed

tests/test_context.py
```diff
@@ -139,6 +140,37 @@ class TrackMe:
...
+        cvar = contextvars.ContextVar('cvar', default='outer')
+        factory_called_future = None
+
+        def factory():
+            nonlocal factory_called_future
+            factory_called_future = self.loop.create_future()
+            self.loop.call_soon(factory_called_future.set_result, None)
+
+            self.assertEqual(cvar.get(), 'inner')
+            return asyncio.Protocol()
```
fantix:
```python
        factory_called_future = self.loop.create_future()

        def factory():
            try:
                self.assertEqual(cvar.get(), 'inner')
            except Exception as e:
                factory_called_future.set_exception(e)
            else:
                factory_called_future.set_result(None)
            return asyncio.Protocol()
```

tests/test_context.py
```diff
@@ -139,6 +140,37 @@ class TrackMe:
...
+
+            s = socket.socket(socket.AF_INET)
+            with s:
+                s.setblocking(False)
+                await self.loop.sock_connect(s, ('127.0.0.1', port))
+
+            await factory_called_future
+
+            srv.close()
+            await srv.wait_closed()
```
fantix:
```python
            try:
                await factory_called_future
            finally:
                srv.close()
                await srv.wait_closed()
```

1st1:
This particular fix is OK, but we need to fix this for all other "native" callbacks in uvloop in a more systematic way.

fantix:
Merged in #348. Thanks for the PR!

Issue #307, opened 2020-01-07
Get uv_loop_t pointer


ghost:
Hello,

For integration purposes I need to get the current underlying uv_loop_t pointer, preferably as a PyCapsule, from uvloop. I need that pointer for use by my python extension and for it to seamlessly integrate with asyncio scripts.

Do you think you could add such a function, or point me in the direction where I can add it? It should be a minimal change.

alexhultman:
I have experimented with passing a Loop object to my extension, accessing the PyTypeObject, confirming the tp_name and tp_itemsize.

From there I planned on accessing the uv_loop_t pointer from the PyObject (the loop pointer looks to be at the top of the object?) but I don't think I have the correct pointer yet. I'm not so sure about how this Cython language works.

Any hints?

1st1:
You can probably add a getter like this in `loop.pyx`:

```
    def get_uvloop_ptr(self):
        return <uint64_t>self.uvloop
```

You can then call it and cast the returned int to a pointer.

alexhultman:
Thanks, would you accept that change?

pranavtbhat:
I've been looking for this functionality too! @alexhultman I can pass in a PR for this if you'd like

alexhultman:
Please make it PyCapsule_New(ptr, null, null) instead so it becomes a proper thing.

https://docs.python.org/3/c-api/capsule.html

Python capsules are like Extern in V8, they are objects that hold raw pointers.

alexhultman:
def get_uvloop_ptr(self):
        return PyCapsule_New(<void *>self.uvloop, NULL, NULL)

alexhultman:
I actually end up doing this:

`asyncio.set_event_loop(asyncio.SelectorEventLoop(uWS.Selector()))`

where `uWS.Selector()` is a minimal object wrapping my internal libuv implementation.

This works for me, but it could still be useful to also have uvloop.get_loop_ptr()

fantix:
Fixed in #310

PR #310, opened 2020-01-17
Expose uv_loop_t pointer for integration with other C-extensions

pranavtbhat:
This PR adds a couple of functions to expose the raw uvloop pointer as int/pycapsule.  #307 

I can remove the int one if we feel it's not required.


alexhultman:
uint64_t is not guaranteed to be the size of void *, why not use the dedicated feature and get it right with one single function?

pranavtbhat:
@1st1  Your thoughts on pycapsule vs int?

pranavtbhat:
The build failure looks unrelated.

XeCycle:
My +1 on this feature; regarding `uint64_t`, `uintptr_t` may be better.  FWIW there are existing libraries exposing pointers as integers (e.g. `numpy.ndarray.ctypes.data`); integers are [easier to work with when using cffi](https://stackoverflow.com/a/33372580).

jensbjorgensen:
Hello, I'm interested in this feature and actually was checking outstanding pull requests before submitting my own. Looking over the above the failed check seems unrelated but as reviewer noted the integer should be uintptr_t not uint64_t. I'm not sure how/if it's possible for me to patch the submitted pull request. Should I submit my own?

fantix:
Thanks for the PR and sorry for the delay! I think having an int pointer in Python doesn't make much sense - it'll be used in a C context anyways, so I think it's fine to only go with the standard API. (I was [suggesting](https://github.com/MagicStack/uvloop/pull/474#discussion_r923685909) the Cython API in another PR, but I like the capsule better - just importing is sufficient, no linking needed!)

I pushed a fix with a Cython test that actually checks the encapsulated pointers.

fantix (approved):

pwuertz:
Two questions on this

- Is there a new `uvloop` release on the horizon that includes this feature? ;)

- Any words of caution concerning `libuv` API compatibility? Is `uvloop` using a statically linked `libuv` library or a shared library that we'd re-use in a c-extension? Does `uvloop` use a vanilla `libuv` for which we can expect SemVer ABI stability?

fantix:
> * Is there a new `uvloop` release on the horizon that includes this feature? ;)

Yes, we'll try to release a new version in the next week or two.

> * Any words of caution concerning `libuv` API compatibility?

Hmm we should probably re-export some of [the libuv version-checking utils](https://docs.libuv.org/en/v1.x/version.html) if we're statically linking, @1st1 wdyt?

> Is `uvloop` using a statically linked `libuv` library or a shared library that we'd re-use in a c-extension? 

This depends, by default it is statically linked but this can be override with the `--use-system-libuv` option.

> Does `uvloop` use a vanilla `libuv` for which we can expect SemVer ABI stability?

Yes, https://github.com/MagicStack/uvloop/tree/master/vendor

fantix:
> > * Any words of caution concerning `libuv` API compatibility?
> 
> Hmm we should probably re-export some of [the libuv version-checking utils](https://docs.libuv.org/en/v1.x/version.html) if we're statically linking, @1st1 wdyt?

`uv_version()` is now exposed in #491.

PR #334, opened 2020-04-11
Add name keyword argument to loop.create_task()

### LINKED ISSUE #309: BUG: uvloop.loop.create_task() is missing 'name' parameter
compare:

1. `uvloop.loop.create_task()`
https://github.com/MagicStack/uvloop/blob/master/uvloop/loop.pyx#L1389

with:

2. `asyncio.loop.create_task()`
https://docs.python.org/3/library/asyncio-eventloop.html?#asyncio.loop.create_task

fantix:
* Fixes #309.
* Refs bpo-34270.

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1386,16 +1386,29 @@ cdef class Loop:
         """Create a Future object attached to the loop."""
         return self._new_future()
 
-    def create_task(self, coro):
+    def create_task(self, coro, *, name=None):
         """Schedule a coroutine object.
 
         Return a task object.
+
+        If name is not None, task.set_name(name) will be called if the task
+        object has the set_name attribute, true for default Task in Python 3.8.
         """
         self._check_closed()
         if self._task_factory is None:
             task = aio_Task(coro, loop=self)
         else:
             task = self._task_factory(self, coro)
+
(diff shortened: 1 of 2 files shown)


uvloop/loop.pyx
```diff
@@ -1386,16 +1386,29 @@ cdef class Loop:
...
         self._check_closed()
         if self._task_factory is None:
             task = aio_Task(coro, loop=self)
         else:
             task = self._task_factory(self, coro)
+
+        # copied from asyncio.tasks._set_task_name (bpo-34270)
+        if name is not None:
+            try:
+                set_name = task.set_name
```
jlaine, line 1406: Does this gain us something over the slightly more readable:

```
try:
    task.set_name(name)
except AttributeError:
    pass
```

1st1: It won't mask an AttributeError coming from the `set_name` implementation. Which is unlikely but can still happen especially with custom task factories.

fantix: Thanks for the suggestion and comments!

I would write something like this:

```python
if name is not None and hasattr(task, "set_name"):
    task.set_name(name)
```

But this is getting into details (`try..except` overhead? Two `O(1)` operations?) and stepping away from the copied code, so I think I'll merge as it is now.

PR #337, opened 2020-04-24
Don't stop the loop in `run_until_complete()` on `SystemExit` and `KeyboardInterrupt` exceptions

### LINKED ISSUE #336: `loop.run_until_complete` raises `RuntimeError` after catching `SystemExit`
* **uvloop version**: 0.14.0
* **Python version**: 3.8.2
* **Platform**: Ubuntu 18.04
* **Does uvloop behave differently from vanilla asyncio? How?**: Yes, vanilla `asyncio` doesn't raise for this code

**Reproduction code:**
```py
import asyncio
import sys
import uvloop

loop = asyncio.new_event_loop()

async def x():
    sys.exit(1)

return_code = 0
try:
    loop.run_until_complete(x())
except SystemExit as exc:
    return_code = exc.code
finally:
    loop.run_until_complete(asyncio.sleep(1))
    sys.exit(return_code)
```
**Traceback:**
```
  File "test.py", line 17, in <module>
    loop.run_until_complete(asyncio.sleep(1))
  File "uvloop/loop.pyx", line 1454, in uvloop.loop.Loop.run_until_complete
RuntimeError: Event loop stopped before Future completed.
```

Same happens when `KeyboardInterrupt` is raised - vanilla `asyncio` handles both of these here: https://github.com/python/cpython/blob/master/Lib/asyncio/base_events.py#L187-L190

jack1142:
Fixes #336

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -361,6 +361,31 @@ class _TestBase:
         # done-callback for the previous future.
         self.loop.run_until_complete(foo(0.2))
 
+    def test_run_until_complete_keyboard_interrupt(self):
+        # Issue #336: run_until_complete() must not schedule a pending
+        # call to stop() if the future raised a KeyboardInterrupt
+        async def raise_keyboard_interrupt():
+            raise KeyboardInterrupt
+
+        self.loop._process_events = mock.Mock()
+
+        try:
+            self.loop.run_until_complete(raise_keyboard_interrupt())
+        except KeyboardInterrupt:
+            pass
+
+        def func():
+            self.loop.stop()
+            func.called = True
+
+        func.called = False
+        try:
+            self.loop.call_soon(func)
+            self.loop.run_forever()
+        except KeyboardInterrupt:
+            pass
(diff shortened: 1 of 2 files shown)


1st1:
Can you add a unittest?

jack1142:
@1st1 I added the unit test based on [CPython's equivalent](https://github.com/python/cpython/blob/003708bcf8f2c58d4b65f68318acf164d713e008/Lib/test/test_asyncio/test_base_events.py#L850-L872) and rebased the PR's branch on top of master.
Let me know if there's anything else you would want me to do.

jack1142:
I've rebased this on top of `master` since it looks like there's a new required check + it's been moved to the project board it seems. I'm still available to do any additional changes if needed so just let me know if needed.

I'm also mentioning this since I'm hoping this could get included in the next release, whenever it's ready of course :)

fantix:
Thanks for the PR! The thing is the test is NOT failing without the fix, I'm looking into it now.

fantix (approved):

Issue #317, opened 2020-02-20
"Assertion `loop->watchers[w->fd] == w' failed." with multiprocessing and OS pipes


momocow:
Hi, I encountered the libuv assertion error while running multiprocessing and communicating with child processes via OS pipes.

I've prepared a minimal reproducible case.

I'm using [kchmck/aiopipe](https://github.com/kchmck/aiopipe) to create pipes, which actually wraps file descriptors from `os.pipe()` into asyncio `StreamReader`s/`StreamWrite`s via `loop.connect_read_pipe()` and `loop.connect_write_pipe()`.

Here, I have decapsulated wrapper classes from `aiopipe` into functions in the `aiopipe_decap.py` for easier code tracing.

```py
# aiopipe_decap.py

import asyncio
import os
from asyncio import (BaseTransport, StreamReader, StreamReaderProtocol,
                     StreamWriter, get_running_loop)
from contextlib import asynccontextmanager, contextmanager


async def _open_reader(fd):
    rx = StreamReader()
    transport, _ = await get_running_loop().connect_read_pipe(
        lambda: StreamReaderProtocol(rx),
        os.fdopen(fd))
    return transport, rx


async def _open_writer(fd):
    rx = StreamReader()
    transport, proto = await get_running_loop().connect_write_pipe(
        lambda: StreamReaderProtocol(rx),
        os.fdopen(fd, "w"))
    tx = StreamWriter(transport, proto, rx, get_running_loop())
    return transport, tx


@asynccontextmanager
async def open_stream(fd, mode):
    transport, stream = await _open_reader(fd) if mode == "r" \
        else await _open_writer(fd)

    try:
        yield stream
    finally:
        try:
            transport.close()
        except OSError:
            # The transport/protocol sometimes closes the fd before this is reached.
            pass

        # Allow event loop callbacks to run and handle closed transport.
        await asyncio.sleep(0)

@contextmanager
def detach(fd):
    os.set_inheritable(fd, True)
    try:
        yield
    finally:
        os.close(fd)

```

And here is the reproducible snippet, which creates a pipe connection between a child process and the master process. Once the child starts up, it will start a loop to send messages back to the master.

You can use the following environment variables to control some of the behaviors in this snippet.
- `U`: enable `uvloop` if this variable is set (default: `False`)
- `R`: number of messages to send (default: `1`)
- `M`: content of messages (default: `"a"`)

```py
# uv.py

import asyncio
import os
from multiprocessing import Process

[...]

PR #342, opened 2020-05-09
forcely add UV_HANDLE_READABLE on pipe_t

### LINKED ISSUE #311: Assertion failure in uv__io_stop
* **uvloop version**: uvloop==0.14.0
* **Python version**: Python 3.7.5 (3.7.5-2~19.10)
* **Platform**: Ubuntu 19.10
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: Yes.
* **Does uvloop behave differently from vanilla asyncio? How?**: Vanilla asyncio works fine.

The attached testcase [test.zip](https://github.com/MagicStack/uvloop/files/4096079/test.zip) crashes uvloop with the following message:
```
python: src/unix/core.c:932: uv__io_stop: Assertion `loop->watchers[w->fd] == w' failed.
```

This looks similar to #88. Here is a stacktrace using gdb: [stacktrace.txt](https://github.com/MagicStack/uvloop/files/4096086/stacktrace.txt)

The real-world application tries to read kadmin’s stdout and write to its stdin. The included mock.py replaces kadmin, but – for this issue – behaves the same. Vanilla asyncio works exactly as expected.

edit: Upon further investigation replacing the two lines
```python
proc.stdin.write (password.encode ('utf-8'))
proc.stdin.write (b'\n')
```

with a single write

```python
proc.stdin.write (password.encode ('utf-8') + b'\n')
```
solves the issue, which indicates a race condition somewhere inside uvloop.

### LINKED ISSUE #312: create_subprocess_exec stdin writer does not write everything

fantix:
* in order to detect peer close on O_WRONLY pipe_t
* partially reverted d8fe153
* refs libuv/libuv#2058
* refs #317
* fixes #311, fixes #312, closes #313

### DIFF
--- a/uvloop/handles/pipe.pyx
+++ b/uvloop/handles/pipe.pyx
@@ -12,6 +12,9 @@ cdef __pipe_init_uv_handle(UVStream handle, Loop loop):
     err = uv.uv_pipe_init(handle._loop.uvloop,
                           <uv.uv_pipe_t*>handle._handle,
                           0)
+    # UV_HANDLE_READABLE allows calling uv_read_start() on this pipe
+    # even if it is O_WRONLY, see also #317, libuv/libuv#2058
+    handle._handle.flags |= uv.UV_INTERNAL_HANDLE_READABLE
     if err < 0:
         handle._abort_init()
         raise convert_error(err)
@@ -147,10 +150,6 @@ cdef class ReadUnixTransport(UVStream):
 @cython.no_gc_clear
 cdef class WriteUnixTransport(UVStream):
 
-    def __cinit__(self):
-        self.disconnect_listener_inited = False
-        self.disconnect_listener.data = NULL
-
     @staticmethod
     cdef WriteUnixTransport new(Loop loop, object protocol, Server server,
                                 object waiter):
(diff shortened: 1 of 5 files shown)


uvloop/handles/pipe.pyx
```diff
@@ -12,6 +12,9 @@ cdef __pipe_init_uv_handle(UVStream handle, Loop loop):
     err = uv.uv_pipe_init(handle._loop.uvloop,
                           <uv.uv_pipe_t*>handle._handle,
                           0)
+    # UV_HANDLE_READABLE allows calling uv_read_start() on this pipe
+    # even if it is O_WRONLY, see also #317, libuv/libuv#2058
+    handle._handle.flags |= 0x00004000  # UV_HANDLE_READABLE
```
1st1, line 17: Can you import the `UV_HANDLE_READABLE` constant via `uv.pxd`?

fantix: Oh sure! I can move the magic number into `uv.pxd` - but it'll be still a magic number in uvloop, because it's an internal flag in libuv, we cannot import it (or maybe I missed anything?).

1st1: Ah, I see. I'd then still add it to `uv.pxd` and maybe name it something like `UV_INTERNAL_HANDLE_READABLE` and add a nice comment to it.

1st1:
Great job, Fantix.

tardyp:
cool! appreciate you included my commit!

PR #353, opened 2020-07-09
Implement loop.shutdown_default_executor()

### LINKED ISSUE #349: Implement shutdown_default_executor()
* **uvloop version**: 0.14.0
* **Python version**: 3.9.0b1
* **Platform**: Linux
* **Does uvloop behave differently from vanilla asyncio? How?**: yes – vanilla asyncio implements this method

Python 3.9 added a new mandatory method to the `AbstractEventLoop` interface, named [shutdown_default_executor()](https://docs.python.org/3.9/library/asyncio-eventloop.html#asyncio.loop.shutdown_default_executor). This method is crucial for `asyncio.run()` to work. If it's absent, `asyncio.run()` will fail with a `NotImplementedError` when performing the shutdown cleanup.

Related CPython issue: https://bugs.python.org/issue34037

aeros:
Closes #349.

Since I don't have too much experience with Cython, I wasn't entirely certain about the purpose of `uvloop/includes/stdlib.pxi`, but I assumed that `threading.Thread` should be included there since it was also done for `concurrent.futures.ThreadPoolExecutor`. At a glance, it seems to be a means of importing the classes from stdlib modules at compile time rather than at run time? I'd appreciate clarification on this if anyone has the time to explain.

Otherwise, it's mostly similar to the reference implementation in [asyncio](https://github.com/python/cpython/pull/15735/files).

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -689,6 +689,17 @@ class _TestBase:
             asyncio.wait_for(foo(), timeout=float('inf')))
         self.assertEqual(res, 123)
 
+    def test_shutdown_default_executor(self):
+        if not hasattr(self.loop, "shutdown_default_executor"):
+            raise unittest.SkipTest()
+
+        async def foo():
+            await self.loop.run_in_executor(None, time.sleep, .1)
+
+        self.loop.run_until_complete(foo())
+        self.loop.run_until_complete(
+            self.loop.shutdown_default_executor())
+
 
 class TestBaseUV(_TestBase, UVTestCase):
 
--- a/uvloop/includes/stdlib.pxi
+++ b/uvloop/includes/stdlib.pxi
@@ -134,6 +134,7 @@ cdef int ssl_SSL_ERROR_WANT_WRITE = ssl.SSL_ERROR_WANT_WRITE
 cdef int ssl_SSL_ERROR_SYSCALL = ssl.SSL_ERROR_SYSCALL
 
 cdef uint64_t MAIN_THREAD_ID = <uint64_t><int64_t>threading.main_thread().ident
+cdef threading_Thread = threading.Thread
 
 cdef int subprocess_PIPE = subprocess.PIPE
 cdef int subprocess_STDOUT = subprocess.STDOUT
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -186,6 +186,8 @@ cdef class Loop:
 
         # Set to True when `loop.shutdown_asyncgens` is called.
         self._asyncgens_shutdown_called = False
+        # Set to True when `loop.shutdown_default_executor` is called.
+        self._executor_shutdown_called = False
 
         self._servers = set()
 
@@ -591,6 +593,7 @@ cdef class Loop:
         self.handler_idle = None
         self.handler_check__exec_writes = None
 
+        self._executor_shutdown_called = True
         executor = self._default_executor
         if executor is not None:
             self._default_executor = None
@@ -2669,6 +2672,8 @@ cdef class Loop:
 
         if executor is None:
             executor = self._default_executor
+            # Only check when the default executor is being used
+            self._check_default_executor()
             if executor is None:
                 executor = cc_ThreadPoolExecutor()
                 self._default_executor = executor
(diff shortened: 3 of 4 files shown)


aeros:
The MacOS CI failures seem to be build-related and not due to the PR. I'm seeing similar CI failures in the [most recent commit](https://github.com/MagicStack/uvloop/commit/465717fdaf2cb260a1c4b230925a537afc22cedb).

uvloop/includes/stdlib.pxi
```diff
@@ -134,6 +134,7 @@ cdef int ssl_SSL_ERROR_WANT_WRITE = ssl.SSL_ERROR_WANT_WRITE
 cdef int ssl_SSL_ERROR_SYSCALL = ssl.SSL_ERROR_SYSCALL
 
 cdef uint64_t MAIN_THREAD_ID = <uint64_t><int64_t>threading.main_thread().ident
+cdef cc_Thread = threading.Thread
```
1st1, line 137: Please rename to `threading_Thread`

uvloop/loop.pyx
```diff
@@ -186,6 +186,8 @@ cdef class Loop:
 
         # Set to True when `loop.shutdown_asyncgens` is called.
         self._asyncgens_shutdown_called = False
+        # Set to True when `loop.shutdown_default_executor` is called.
+        self._executor_shutdown_called = False
```
1st1, line 190: I think you should also add the `_executor_shutdown_called` field to loop.pxd.

uvloop/includes/stdlib.pxi
aeros: Oh, I just realized that `cc` was short for *concurrent* in the context of `cc_ThreadPoolExecutor`, haha. Didn't catch that earlier since I usually see `concurrent.futures` abbreviated as `cf`.

aeros:
I'll wait on squashing into a single commit until the review process is complete, since it tends to affect previous review comments after force-pushing to the PR branch.

uvloop/loop.pyx
aeros: Yep, that would make sense, seeing as `_asyncgens_shutdown_called` is also present there. I'm just not entirely used to working with languages that use structs to specify the fields in header files. :-)

tests/test_base.py
```diff
@@ -689,6 +689,17 @@ async def foo():
...
+    def test_shutdown_default_executor(self):
+        if not hasattr(self.loop, "shutdown_default_executor"):
+            raise unittest.SkipTest()
+
+        async def foo():
+            await self.loop.run_in_executor(None, time.sleep, .1)
+
+        self.loop.run_until_complete(foo())
+        self.loop.run_until_complete(
+            self.loop.shutdown_default_executor())
```
1st1, line 701: Is this the only test we have in cpython/asyncio/tests?

uvloop/loop.pyx
1st1: Yeah, Cython looks a bit alien at first

tests/test_base.py
aeros: Actually, it looks like [this](https://github.com/python/cpython/blob/9650fe0197779b4dfded94be111e39c5810f098f/Lib/test/test_asyncio/test_events.py#L358-L376) is the only test that uses it in CPython:

```py
    def test_run_in_executor_cancel(self):
        called = False

        def patched_call_soon(*args):
            nonlocal called
            called = True

        def run():
            time.sleep(0.05)

        f2 = self.loop.run_in_executor(None, run)
        f2.cancel()
        self.loop.run_until_complete(
                self.loop.shutdown_default_executor())
        self.loop.close()
        self.loop.call_soon = patched_call_soon
        self.loop.call_soon_threadsafe = patched_call_soon
        time.sleep(0.4)
        self.assertFalse(called)
```

In retrospect, I probably should have added some dedicated test(s) for `shutdown_default_executor()` in the PR that added it in the first place, but at the time, I didn't have a decent understanding of asyncio unit tests. Should I add this one to the CPython regression tests and maybe a couple more if I can think of some that could be useful?

1st1: > Should I add this one to the CPython regression tests and maybe a couple more if I can think of some that could be useful?

The more tests the better :) If they are not redundant, of course. I'll merge this PR now, huge thanks!

qeternity:
Over 4 months since this was merged - is there any hope for a new pypi release?

1st1:
In the next couple of weeks.

qeternity:
@1st1 not to be a nag but anything that the community can do in terms of preflight checks?

Issue #358, opened 2020-08-13
mypy show type hint error


kulame:
i use mypy to check my code.
but it metioned:
    error: Skipping analyzing 'uvloop': found module but no type hints or library stubs

does uvloop support mypy check?

1st1:
Maybe there's a typeshed for it, but there's no out of the box support. You're welcome to work on adding that!

fantix:
This is fixed in #366 I think.

PR #366, opened 2020-11-02
Add typings

bryanforbes:
This PR adds typings for use with type checkers like `mypy`. Overall, this was fairly easy but two errors have to be suppressed with a `# type: ignore` comment:

* `remove_signal_handler()` returns a `bool` and typeshed has this method returning `None` in `AbstractEventLoop`. This seems to be an error in typeshed since the [docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.remove_signal_handler) state it should return a bool.
* `create_connection()` in Python 3.8+ accepts two new arguments (`happy_eyeballs_delay` and `interleave`) and since `uvloop.Loop.create_connection()` doesn't allow those arguments, the two methods are incompatible. This may be unavoidable.

References #358

### DIFF
--- a/uvloop/loop.pyi
+++ b/uvloop/loop.pyi
@@ -1,6 +1,5 @@
 import asyncio
 import ssl
-import sys
 from socket import AddressFamily, SocketKind, _Address, _RetAddress, socket
 from typing import (
     IO,
--- a/uvloop/loop.pyi
+++ b/uvloop/loop.pyi
@@ -1,0 +1,290 @@
+import asyncio
+import ssl
+import sys
+from socket import AddressFamily, SocketKind, _Address, _RetAddress, socket
+from typing import (
+    IO,
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    TypeVar,
+    Union,
+    overload,
+)
+
+_T = TypeVar('_T')
+_Context = Dict[str, Any]
+_ExceptionHandler = Callable[[asyncio.AbstractEventLoop, _Context], Any]
+_SSLContext = Union[bool, None, ssl.SSLContext]
+
+class Loop:
+    def call_soon(
+        self, callback: Callable[..., Any], *args: Any, context: Optional[Any] = ...
+    ) -> asyncio.Handle: ...
+    def call_soon_threadsafe(
+        self, callback: Callable[..., Any], *args: Any, context: Optional[Any] = ...
+    ) -> asyncio.Handle: ...
+    def call_later(
+        self, delay: float, callback: Callable[..., Any], *args: Any, context: Optional[Any] = ...
+    ) -> asyncio.TimerHandle: ...
+    def call_at(
+        self, when: float, callback: Callable[..., Any], *args: Any, context: Optional[Any] = ...
+    ) -> asyncio.TimerHandle: ...
+    def time(self) -> float: ...
+    def stop(self) -> None: ...
+    def run_forever(self) -> None: ...
+    def close(self) -> None: ...
+    def get_debug(self) -> bool: ...
+    def set_debug(self, enabled: bool) -> None: ...
+    def is_running(self) -> bool: ...
+    def is_closed(self) -> bool: ...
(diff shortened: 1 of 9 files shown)


uvloop/loop.pyi
```diff
@@ -0,0 +1,397 @@
...
+    def set_read_buffer_limits(self, high: Optional[int] = ..., low: Optional[int] = ...) -> None: ...
+    def get_read_buffer_limits(self) -> Tuple[int, int]: ...
+    def get_read_buffer_size(self) -> int: ...
+    def write(self, data: Any) -> None: ...
+    def writelines(self, list_of_data: List[Any]) -> None: ...
+    def write_eof(self) -> None: ...
+    def can_write_eof(self) -> bool: ...
+    def abort(self) -> None: ...
+
+class _Server(asyncio.AbstractServer):
```
1st1, line 110: I'm curious why we need this as opposed to simply using `asyncio.AbstractServer`?

bryanforbes: `asyncio.AbstractServer` is defined as follows in typeshed:

```python
class AbstractServer:
    sockets: Optional[List[socket]]
    def close(self) -> None: ...
    if sys.version_info >= (3, 7):
        async def __aenter__(self: _T) -> _T: ...
        async def __aexit__(self, *exc: Any) -> None: ...
        def get_loop(self) -> AbstractEventLoop: ...
        def is_serving(self) -> bool: ...
        async def start_serving(self) -> None: ...
        async def serve_forever(self) -> None: ...
    async def wait_closed(self) -> None: ...
```

Since `uvloop` works on Python 3.5+ and `Server` defines those methods for all versions, `asyncio.AbstractServer` will be wrong for `uvloop` on Python 3.5 and 3.6.

1st1: I'll be dropping support for 3.5 and 3.6. 3.5 is no longer supported by Python core devs; 3.6 is in security fixes only mode. And with 3.9 now out, uvloop will be 3.7+ only in its next release, supporting three Python versions: 3.7, 3.8, and 3.9.

bryanforbes: That's good to know! I'll go back over this PR with that in mind.

bryanforbes: One thing to note: I think we'll still need `_Server` because `AbstractServer.sockets` is `Optional[List[Socket]]` and `Server.sockets` is `List[Socket]`.

1st1: I think we should stick to AbstractServer typing here too. That makes it easy for us to fix our `Server.sockets` property to follow the spec.

bryanforbes: Some of these others I added while I was navigating the source. I'll make sure I validate all of these are needed when I go over this again.

1st1: Yes, thanks. And we can always fix uvloop code that doesn't conform to asyncio APIs.

bryanforbes: Sorry that it's taken so long to get back to this. I've updated the PR based on your feedback and rebased on master. Unfortunately, the PR checks are failing prematurely.

1st1:
@fantix We can probably merge this one too, please review.

bryanforbes:
I can rebase this on master if it would make it easier to review. Just let me know.

fantix:
@bryanforbes Yes, please! Thank you!

bryanforbes:
I have a couple of small edits to make before you merge

fantix (approved):
LGTM!

bryanforbes:
@fantix I committed and pushed the edits. Let me know if you have questions about them. One thing to note: the two things I pointed out in the description of this PR are still issues. Once those methods are changed, the `# type: ignore[misc]` can be removed from `__init__.py`.

fantix:
> Once those methods are changed, the `# type: ignore[misc]` can be removed from `__init__.py`.

Sounds good 👍

1st1:
Thanks @bryanforbes!

bryanforbes:
Not a problem! If you ever have any typing questions, feel free to reach out to me.

1st1:
@bryanforbes thanks! If you have time please take a look at https://github.com/MagicStack/immutables/pull/54 -- another our package.

Issue #367, opened 2020-11-02
Wheel support for linux aarch64


odidev:
**Summary**
Installing uvloop on aarch64 via pip using command "pip3 install uvloop" tries to build wheel from source code.

**Problem description**
uvloop doesn't have wheel for aarch64 on PyPI repository. So, while installing uvloop via pip on aarch64, pip builds wheel for same resulting in it takes more time to install uvloop. Making wheel available for aarch64 will benefit aarch64 users by minimizing uvloop installation time.

**Expected Output**
Pip should be able to download uvloop wheel from PyPI repository rather than building it from source code.

@uvloop-team, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make uvloop wheel available for aarch64. It will be a great opportunity for me to work with you.

elprans:
Added in #376.  Please note that manylinux on aarch64 is [not stable yet](https://github.com/pypa/manylinux/issues/84), so consider aarch64 wheels to be beta.

PR #369, opened 2020-11-18
Add wheel support for aarch64

odidev:
Added wheel support for aarch64.
Related to #367 .


1st1:
@elprans would appreciate you taking a look

ambv:
We should sit on this until aarch64 is fully specced out by the PyPA. More context at https://github.com/hynek/argon2-cffi/pull/74

elprans:
The current Github Actions workflow (merged in #376) contains support for aarch64 and 0.15.0 was actually released with aarch64 wheels.  Lets see how that works out.  If people start reporting weird issues on ARM, we'll pull them down.

PR #374, opened 2020-12-12
Use full path to performance.png

asfaltboy:
To make it visible in https://pypi.org/project/uvloop/


1st1 (approved):
It's a bit unfortunate that we have to do this :/

PR #399, opened 2021-02-13
Fix ref issue when protocol is in Cython

fantix:
Because `context.run()` doesn't hold reference to the callable, when e.g. the protocol is written in Cython, the callbacks were not guaranteed to hold the protocol reference. This PR fixes the issue by explicitly add a reference before `context.run()` calls.

Refs edgedb/edgedb#2222

### DIFF
--- a/uvloop/handles/basetransport.pyx
+++ b/uvloop/handles/basetransport.pyx
@@ -70,7 +70,9 @@ cdef class UVBaseTransport(UVSocketHandle):
             try:
                 # _maybe_pause_protocol() is always triggered from user-calls,
                 # so we must copy the context to avoid entering context twice
-                self.context.copy().run(self._protocol.pause_writing)
+                run_in_context(
+                    self.context.copy(), self._protocol.pause_writing,
+                )
             except (KeyboardInterrupt, SystemExit):
                 raise
             except BaseException as exc:
@@ -91,7 +93,9 @@ cdef class UVBaseTransport(UVSocketHandle):
                 # We're copying the context to avoid entering context twice,
                 # even though it's not always necessary to copy - it's easier
                 # to copy here than passing down a copied context.
-                self.context.copy().run(self._protocol.resume_writing)
+                run_in_context(
+                    self.context.copy(), self._protocol.resume_writing,
+                )
             except (KeyboardInterrupt, SystemExit):
                 raise
             except BaseException as exc:
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -89,6 +89,34 @@ cdef inline socket_dec_io_ref(sock):
         sock._decref_socketios()
 
 
+cdef inline run_in_context(context, method):
+    # This method is internally used to workaround a reference issue that in
+    # certain circumstances, inlined context.run() will not hold a reference to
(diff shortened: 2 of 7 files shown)


1st1:
> Because context.run() doesn't hold reference to the callable

Perhaps we should also fix this in CPython 3.9.

uvloop/handles/basetransport.pyx
```diff
@@ -88,10 +90,12 @@ cdef class UVBaseTransport(UVSocketHandle):
         if self._protocol_paused and size <= self._low_water:
             self._protocol_paused = 0
             try:
+                # We need to hold a reference here because context.run() won't
+                resume_writing = self._protocol.resume_writing
                 # We're copying the context to avoid entering context twice,
                 # even though it's not always necessary to copy - it's easier
                 # to copy here than passing down a copied context.
-                self.context.copy().run(self._protocol.resume_writing)
+                self.context.copy().run(resume_writing)
```
1st1, line 98: I propose to do this differently. Create a helper function "run_in_context" and use it like

```
run_in_context(
    self.context.copy(),
    self._protocol.resume_writing
)
```

1st1: That new function then should consistently be used across the uvloop codebade.

uvloop/loop.pyx
```diff
@@ -89,6 +89,22 @@ cdef inline socket_dec_io_ref(sock):
         sock._decref_socketios()
 
 
+cdef inline run_in_context(context, method):
+    # This method is internally used to workaround a reference issue that in
```
1st1, line 93: My understanding here is that `method` is going to be increfed by cython even if `run_in_context` is inlined? Should we add explicit INCREF/DECREF here for clarity?

fantix: > `method` is going to be increfed by cython even if `run_in_context` is inlined?

Yes, Cython is always creating a new reference and `GOTREF` before calling the inlined function, and `DECREF` afterwards, unless the `method` is already a caller's local reference.

> Should we add explicit INCREF/DECREF here for clarity?

I'm okay either way - as this is likely a temporary workaround depending on if we should fix Cython or CPython, upon which I'd appreciate some light shed.

1st1: Let's add explicit incref/decref, the code will be way more clear that way. The performance penalty is going to be negligible. While it's temporary, we don't know how temporary it will be. If we want to fix Cython it can be an issue for a few months (although this doesn't look like a Cython issue, after all, right?). And if we want to fix CPython, it will take years before we stop supporting 3.9.1.

uvloop/loop.pyx
```diff
@@ -89,6 +89,22 @@ cdef inline socket_dec_io_ref(sock):
...
+    # certain circumstances, inlined context.run() will not hold a reference to
+    # the given method instance, which - if deallocated - will cause segault.
+    # See also: edgedb/edgedb#2222
+    return context.run(method)
```
1st1, line 97:
```suggestion
    INCREF(method)
    try:
        return context.run(method)
    finally:
        DECREF(method)
```

(this is what I meant, just in case)

PR #407, opened 2021-04-12
fix  `call_soon_threadsafe` but not really thread-safe

hehaha:


hehaha:
link with the [issue](https://github.com/MagicStack/uvloop/issues/408)

PR #428, opened 2021-07-30
add support for <timer handle>.when()

jensbjorgensen:
here's a simple implementation to add support for when() on a timer handle

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -852,6 +852,15 @@ class TestBaseUV(_TestBase, UVTestCase):
         self.assertEqual(OK, 5)
         self.assertEqual(NOT_OK, 0)
 
+    def test_loop_call_later_handle_when(self):
+        cb = lambda: False  # NoQA
+        delay = 1.0
+        loop_t = self.loop.time()
+        handle = self.loop.call_later(delay, cb)
+        self.assertAlmostEqual(handle.when(), loop_t + delay, places=2)
+        handle.cancel()
+        self.assertTrue(handle.cancelled())
+
 
 class TestBaseAIO(_TestBase, AIOTestCase):
     pass
--- a/uvloop/cbhandles.pyx
+++ b/uvloop/cbhandles.pyx
@@ -308,6 +308,9 @@ cdef class TimerHandle:
     def cancel(self):
         self._cancel()
 
+    def when(self):
+        return self.timer.get_when() * 1e-3
+
 
 cdef format_callback_name(func):
     if hasattr(func, '__qualname__'):
--- a/uvloop/handles/timer.pyx
+++ b/uvloop/handles/timer.pyx
@@ -23,6 +23,7 @@ cdef class UVTimer(UVHandle):
         self.ctx = ctx
         self.running = 0
         self.timeout = timeout
+        self.start_t = 0
 
     cdef stop(self):
         cdef int err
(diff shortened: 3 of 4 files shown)


uvloop/handles/timer.pyx
```diff
@@ -41,6 +42,10 @@ cdef class UVTimer(UVHandle):
 
     cdef start(self):
         cdef int err
+        cdef timespec ts
+
+        timespec_get(&ts, 1)
+        self.when = ts.tv_sec + ts.tv_nsec*1e-9 + self.timeout*1e-3
```
1st1, line 48: Thanks for the contribution! I'd make this calculation on demand. Add a method to the Timer handle `get_when()`. Also we need tests.

jensbjorgensen: ok instead of calculating the thole thing there you're thinking I should cache the start time (as double instead of the timespec?) and then calculate off of the self.time in get_when, do I understand correctly?

jensbjorgensen: any update @1st1 ? Just looking for a little guidance before I reform my patch

1st1: Sorry for the delay! My idea was to do as little work as possible in `__cinit__` or `__init__`, offsetting the bulk of the work to `when()`.

jensbjorgensen: I see, so I need to at a minimum keep the current time I've sampled with timespec_get(...), so one option to avoid any calculation would be to just keep the timespec struct. I didn't go that route because it would require twice the storage as sizeof(timespec) == 16 vs. sizeof(self.when) == 8. Buf if it's your preference to cache the bytes and defer the calculation then I'll change it as you say. Is that your preference?

1st1: I think we should call `loop.time()` when the handle is created and store that time in the handle. The method returns monotonic time that's cached for the current iteration and that's what libuv is using internally to compute the timeouts. Then you'd simply need to add the timeout to that captured creation time in `.when()`

1st1: Also using `loop.time()` is the only way to get consistency with `loop.call_at()`.

jensbjorgensen: Ok sure, fair enough. That's certainly consistent with the documentation.

jensbjorgensen: Ok, pushed another commit. Works the way you mention and I added one test, not sure what other sort of testing will be worthwhile.

uvloop/cbhandles.pyx
```diff
@@ -308,6 +308,9 @@ cdef class TimerHandle:
     def cancel(self):
         self._cancel()
 
+    def when(self):
+        return self.timer.get_when()*1e-3
```
1st1, line 312:
```suggestion
        return self.timer.get_when() * 1e-3
```

1st1 (approved):
Great work, please fix the one remaining nit

jensbjorgensen:
ok, did that, so we're there?! :-D

elprans:
Tests seem to be failing in CI: https://github.com/MagicStack/uvloop/pull/428/checks?check_run_id=3292953606#step:7:2062

tests/test_base.py
```diff
@@ -852,6 +852,15 @@ async def main():
         self.assertEqual(OK, 5)
         self.assertEqual(NOT_OK, 0)
 
+    def test_loop_call_later_handle_when(self):
+        cb = lambda: False  # NoQA
+        delay = 0.01
+        loop_t = self.loop.time()
+        handle = self.loop.call_later(delay, cb)
+        self.assertAlmostEqual(handle.when(), loop_t + delay)
```
elprans, line 860:
```suggestion
        self.assertAlmostEqual(handle.when(), loop_t + delay, places=5)
```

PR #442, opened 2021-09-22
setup.py: append environment CFLAGS to local CFLAGS variable

giuliobenetti:
Please check build failure [1] due to toolchain m68k infinite loop if -O2 is passed.
There we need to override it with -O0 and the only way I've found is the on in the patch.

[1]: http://autobuild.buildroot.net/results/17d/17d6e6422abadcd6313c430c40f2a5d7162dbbd3/build-end.log


giuliobenetti:
My bad, this doesn't work, I've tested it badly. I will reopen a PR with a correct solution.

Issue #404, opened 2021-02-28
Using vectorized IO (scatter/gather)


jakirkham:
Operations like `writelines` in [the Stream]( https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamWriter.writelines ) and [the Transport]( https://docs.python.org/3/library/asyncio-protocol.html?highlight=writelines#asyncio.WriteTransport.writelines ) APIs provide library authors the opportunity to send collections of buffers that they would like written, sent, etc. in one go. This can be really handy as it takes only one pass (as opposed to multiple passes) through layers of code to prepare buffers before they go out.

Many OSes supply similar C-level operations like [`writev` on POSIX compatible or similar on Windows]( https://en.wikipedia.org/wiki/Vectored_I/O ) for operating on file descriptors. Similarly [`sendmsg` on Linux and Unix]( https://linux.die.net/man/3/sendmsg ) or [`WSASend` on Windows]( https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasend ) provide implementations for sockets.

Admittedly am not very familiar with `libuv`'s API (so maybe devs here can comment on this), but it appears there are some APIs in `libuv` like [`uv_write`]( http://docs.libuv.org/en/v1.x/stream.html#c.uv_write ) can take multiple buffers, which can [internally redirect]( https://github.com/libuv/libuv/blob/285a5ea819035ff777b8b7c6a367f3f5b55d8809/src/unix/udp.c#L728 ) to [`sendmsg`]( https://github.com/libuv/libuv/blob/285a5ea819035ff777b8b7c6a367f3f5b55d8809/src/unix/udp.c#L448 ) or [`WSASend`]( https://github.com/libuv/libuv/blob/4ddc292774be827a297449e2d5ef4047c85de7ca/src/win/tcp.c#L934 ). This also appears to be true for files with the [`uv_fs_write`]( http://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_write ) API.

[...]

PR #445, opened 2021-10-01
Queue write only after processing all buffers

jakirkham:
Partially addresses ( https://github.com/MagicStack/uvloop/issues/404 )

Delays calling `_queue_write` until `writelines` has added all buffers with `_write`. Should allow `libuv` to leveraged vectorized IO (scatter) via calling `sendmsg` under-the-hood.

### DIFF
--- a/uvloop/handles/stream.pyx
+++ b/uvloop/handles/stream.pyx
@@ -162,7 +162,7 @@ cdef class _StreamWriteContext:
                     PyObject_GetBuffer(
                         buf, &p_pybufs[py_bufs_len], PyBUF_SIMPLE)
                 except Exception:
-                    # This shouldn't ever happen, as `UVStream._write`
+                    # This shouldn't ever happen, as `UVStream._buffer_write`
                     # casts non-bytes objects to `memoryviews`.
                     ctx.py_bufs_len = py_bufs_len
                     ctx.free_bufs()
@@ -407,7 +407,7 @@ cdef class UVStream(UVBaseTransport):
 
         return written
 
-    cdef inline _write(self, object data):
+    cdef inline _buffer_write(self, object data):
         cdef int dlen
 
         if not PyBytes_CheckExact(data):
@@ -420,6 +420,7 @@ cdef class UVStream(UVBaseTransport):
         self._buffer_size += dlen
         self._buffer.append(data)
 
+    cdef inline _write(self):
         if (not self._protocol_paused and
                 (<uv.uv_stream_t*>self._handle).write_queue_size == 0 and
                 self._buffer_size > self._high_water):
@@ -443,9 +444,10 @@ cdef class UVStream(UVBaseTransport):
                 # If not all of the data was sent successfully,
                 # we might need to pause the protocol.
                 self._maybe_pause_protocol()
-            return
 
-        self._maybe_pause_protocol()
+        elif self._buffer_size > 0:
+            self._maybe_pause_protocol()
+            self._loop._queue_write(self)
 
     cdef inline _exec_write(self):
         cdef:
@@ -678,8 +680,8 @@ cdef class UVStream(UVBaseTransport):
         if self._conn_lost:
             self._conn_lost += 1
             return
-        self._write(buf)
-        self._loop._queue_write(self)
+        self._buffer_write(buf)
+        self._write()
 
     def writelines(self, bufs):
         self._ensure_alive()
@@ -690,8 +692,8 @@ cdef class UVStream(UVBaseTransport):
             self._conn_lost += 1
             return
         for buf in bufs:
-            self._write(buf)
-        self._loop._queue_write(self)
+            self._buffer_write(buf)
+        self._write()
 
     def write_eof(self):
         self._ensure_alive()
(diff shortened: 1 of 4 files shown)


uvloop/handles/stream.pyx
```diff
@@ -1,3 +1,8 @@
+# MIT and Apache 2.0
+#
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
jakirkham, line 3: Not sure how noting copyright holders/contributors works here. If there is a preferred way to do this, please let me know and we can handle that way

jakirkham:
Also given GitHub's new policy regarding running CI. Think someone needs to approve running CI here for that to happen

uvloop/handles/stream.pyx
1st1: Where is this copyright coming from?

1st1: Thank you for the contribution. This is the first time a contributor asks to add explicit copyright notices to a file and I don't think I like to add them everywhere, especially for minor contributions.

Would changing "Copyright (c) 2015-present MagicStack Inc.  http://magic.io" to "Copyright (C) 2016-present the uvloop authors and contributors." in our LICENSE files alleviate the need of an explicit copyright note you?

1st1: Also: https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/

jakirkham: Yeah I think that would help if that would be agreeable. Would you like to make that change or should I make that here?

jakirkham: Have pushed the change as suggested to the PR. Please let me know what you think :)

jakirkham:
@1st1 any thoughts here?

Also need some help running the tests. GH doesn't run tests for new contributors on any commit without approval by default

jakirkham:
Friendly nudge 😉

fantix:
Thanks for the PR and sorry for the long waiting!

> Should allow libuv to leveraged vectorized IO (scatter) via calling sendmsg under-the-hood.

Hmm do you have some test results to show if gathered send is/not working before/after this change? I'm asking because I think `_queue_write()` doesn't call `_exec_write()` [until the next loop iteration](http://docs.libuv.org/en/v1.x/check.html), so calling `_queue_write()` multiple times in a tight loop in `writelines()` is equivalent to calling it just once. I'll run a test later when I come back to this one.

1st1 (approved):
@fantix looks good. Please check that we call `_queue_write` whenever we call `_write`, but overall this is good.

jakirkham:
Thanks all! 🙏

Yeah agree benchmarking makes sense. Is there a preferred set of benchmarks here we should try with initially? Also is there somewhere to add new benchmarks?

As to context about this change, did a fair bit of sleuthing in issue ( https://github.com/MagicStack/uvloop/issues/404 ), which included a bit of digging into `libuv` itself. Could always be missing something. So would certainly appreciate feedback on that.

Sure happy to tweak the API to make things a bit more flexible.

uvloop/handles/stream.pyx
```diff
@@ -446,7 +446,6 @@ cdef class UVStream(UVBaseTransport):
             return
 
         self._maybe_pause_protocol()
-        self._loop._queue_write(self)
```
fantix, line 449: This is actually wrong - in `_write()`, we have two early exits that don't start the `UVCheck` handler by calling `loop._queue_write()`. By moving this call out, we always start the `UVCheck` even if there is nothing to send.

Also, because of the fast-path in `_write()`, this doesn't really push all the buffers before trying to send in order to leverage the scatter sending. I'm pushing a new implementation to fix that.

uvloop/handles/stream.pyx
```diff
@@ -420,6 +420,7 @@ cdef class UVStream(UVBaseTransport):
         self._buffer_size += dlen
         self._buffer.append(data)
 
+    cdef inline _write(self):
```
1st1, line 423: I think we should we rename it to `_queue_write()`?

1st1 (approved):

uvloop/handles/stream.pyx
fantix: Good point! Though I would call it `_start_write()`, because:

1. This method may try to write immediately without queuing.
2. There is also `loop._queue_write()` which actually schedules a write in the next iteration.
3. In `UVStream`, we also have `_exec_write()`, `_try_write()`, now with `_buffer_write()`, `_queue_write()` just confuses people a lot - which does what?

So I'm also adding some comments now.

1st1: Awesome, this makes sense! Maybe `_initiate_write()`? Like it's closer to "start now or soon".

fantix:
@jakirkham thank you for starting this anyways! Please feel free to comment on the PR or the original issue.

uvloop/handles/stream.pyx
fantix: Sure thing!

jakirkham:
Thanks all for improving this and including it 🙏

Sorry for being unresponsive (was out at the time). Though am happy to see this conclusion while catching up 🙂

PR #435, opened 2021-08-09
Use cibuildwheel to build wheels

elprans:

PR #447, opened 2021-10-23
Version bump of cibuildwheel

WAKayser:
Update cibuildwheel to 2.2.0, which will enable musllinux wheel builds. This is useful for environments like alpine linux containers.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -108,7 +108,7 @@ jobs:
       run: |
         brew install gnu-sed libtool autoconf automake
 
-    - uses: pypa/cibuildwheel@v2.1.1
+    - uses: pypa/cibuildwheel@v2.2.0
       env:
         CIBW_BUILD_VERBOSITY: 1
         CIBW_BUILD: ${{ matrix.cibw_python }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -108,7 +108,7 @@ jobs:
       run: |
         brew install gnu-sed libtool autoconf automake
 
-    - uses: pypa/cibuildwheel@v2.2.0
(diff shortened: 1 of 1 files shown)


br3ndonland:
Thanks for taking care of this @WAKayser. I landed here after wondering why my uvloop installs were still so slow on Alpine Linux.

Cross-referencing the related PR MagicStack/uvloop#435 reviewed by @fantix. Would appreciate a review here as well.

Happy to help, if there's any way I can help move this forward.

ben9923:
I believe further changes are needed here.

There's currently a workaround for `cffi` wheels, invoking `yum`.
```yaml
# This is needed for now because cffi has no cp310 wheels
CIBW_BEFORE_ALL_LINUX: "yum -y install libffi-devel"
```

Running this command on Alpine would obviously fail, so it must be changed.
One option could simply be checking if `yum` exists as part of the command.
Another option would be using `cibuildwheel`'s override mechanism (toml-only, added quite recently IIRC,):
```toml
[[tool.cibuildwheel.overrides]]
select = "*-manylinux*"
before-all = "yum -y install libffi-devel"
```

While this is actually unnecessary now, as `cffi` has `cp310` wheels, adding `cp311` will re-introduce this requirement.
`cffi` also has no `musllinux` wheels, but it seems like the default image already has `libffi-dev`.

Might also be worth upgrading `cibuildwheel` to the latest version. It will change from `manylinux2010` to `manylinux2014` so if it's not desired an image would need to be set explicitly.

I would love to see musl wheels too, let me know if I can help :)
CC @elprans @fantix

fantix:
Thanks for the PR! This is already done now and musl wheels are built successfully in my latest test, will be included in the next release.

PR #453, opened 2021-12-08
fix incorrect main thread id value in mp.Process 

### LINKED ISSUE #452: incorrect main thread definition in subprocess
Hi, I have program with main process and a bunch subprocesses. Main process sends tasks via pipe to some subprocesses. Each subprocess are getting task synchronously in a separate thread and then process (send http query) in main thread of subprocess. I wanted to make `add_signal_handler` in main thread of each subprocess. But I got `ValueError: add_signal_handler() can only be called from the main thread` with `uvloop`. Asyncio works correctly. I think the main problem in `MAIN_THREAD_ID`. It's initialized only once.

* **uvloop version**: 0.16.0
* **Python version**: 3.8.10
* **Platform**: `Linux devvoip 5.4.0-91-generic #102-Ubuntu SMP Fri Nov 5 16:31:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

* **Does uvloop behave differently from vanilla asyncio? How?**: `uvloop` raises ValueError, `asyncio` works fine

horpto:
fixes #452

MAIN_THREAD_ID had incorrect value when you run uvloop in subprocess. So it was affecting `_setup_or_resume_signals`. Subprocess is runned via `multiprocessing.Process` and should be closed by signal. However subprocess gets signal and stucks in eternal loop.

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -141,7 +141,6 @@ cdef class Loop:
 
         self._closed = 0
         self._debug = 0
-        self._thread_is_main = 0
         self._thread_id = 0
         self._running = 0
         self._stopping = 0
@@ -215,7 +214,7 @@ cdef class Loop:
         self._servers = set()
 
     cdef inline _is_main_thread(self):
-        return MAIN_THREAD_ID == PyThread_get_thread_ident()
+        return threading_main_thread().ident == PyThread_get_thread_ident()
 
     def __init__(self):
         self.set_debug((not sys_ignore_environment
@@ -519,7 +518,6 @@ cdef class Loop:
         self._last_error = None
 
         self._thread_id = PyThread_get_thread_ident()
-        self._thread_is_main = MAIN_THREAD_ID == self._thread_id
         self._running = 1
 
         self.handler_check__exec_writes.start()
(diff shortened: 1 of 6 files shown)


horpto:
@1st1 hello, could you please review or call someone else who can review

uvloop/loop.pyx
```diff
@@ -216,7 +215,7 @@ cdef class Loop:
         self._servers = set()
 
     cdef inline _is_main_thread(self):
-        return MAIN_THREAD_ID == PyThread_get_thread_ident()
+        return threading_main_thread().ident == PyThread_get_thread_ident()
```
1st1, line 218: @fantix this is a pretty performance-critical function. I don't like calling `threading_main_thread()` here. We should instead cache it on the first call.

fantix: I think so too.

Even though the code change makes sense, I'm still trying to reproduce the original issue with `multiprocessing`. It just seems it's a very specific mixing usage of uvloop and `multiprocessing` - I'll add a test if reproduced.

fantix (approved):
Thanks for the PR! I added a test and made the main thread ID a lazy value for runtime performance.

PR #454, opened 2021-12-09
create_subprocess_exec should treat env={} as empty environment (#439)

### LINKED ISSUE #439: create_subprocess_exec treats env={} differently than vanilla asyncio
* **uvloop version**: 0.16.0
* **Python version**: 3.9.6
* **Platform**: macOS 11.5.2
* **Does uvloop behave differently from vanilla asyncio? How?**:  create_subprocess_exec treats `env` differently.

When I set env={}, uvloop treats this as None and inherits all environment variables, while vanilla asyncio treats {} as an empty environment.

```python
import asyncio
import os

import pytest

if os.environ.get("UVLOOP"):
    # Install uvloop event loop.
    import uvloop

    @pytest.fixture
    def event_loop():
        yield loop
        loop.close()


@pytest.mark.asyncio
async def test_env():
    "This test passes under vanilla asyncio but fails under uvloop."
    proc = await asyncio.create_subprocess_exec(
        "/usr/bin/env",
        stdout=asyncio.subprocess.PIPE,
        env={},
    )
    result = await proc.communicate()

    # Under uvloop, stdout contains the full, inherited environment.
    assert result == (b"", None)
```

byllyfish:
Fixes #439

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -49,6 +49,22 @@ class _TestProcess:
 
         self.loop.run_until_complete(test())
 
+    def test_process_env_2(self):
+        async def test():
+            cmd = 'env'
+            env = {}  # empty environment
+            proc = await asyncio.create_subprocess_exec(
+                cmd,
+                env=env,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE)
+
+            out, _ = await proc.communicate()
+            self.assertEqual(out, b'')
+            self.assertEqual(proc.returncode, 0)
+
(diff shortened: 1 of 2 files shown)


fantix:
I'm not sure why the new test is failing on the CI? It works locally for me. Other than that, I think this is okay, thanks for the PR!

1st1 (approved):

1st1:
@fantix we should merge this

fantix:
Dropped an assertion under debug mode that ensures cstring array must not be empty.

PR #455, opened 2021-12-20
Run tests against Python 3.10 stable

michael-k:
`3.10` without quotes is interpreted as `3.1` (float).


fantix:
Fixed in #477, thanks for the PR though!

PR #466, opened 2022-05-11
avoid self._errpipe_write double close

### LINKED ISSUE #414: Pathlib throws "Bad file descriptor", but only with `uvloop`.
* **uvloop version**: 0.15.3
* **Python version**: 3.9.5
* **Platform**: Manjaro, kernel is 5.10.42-1
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: no
* **Does uvloop behave differently from vanilla asyncio? How?**: without `uvloop`, the script works just fine.

When spawning processes and creating file, an `OSError: [Errno 9] Bad file descriptor` sometimes appears.

Script and output: https://gist.github.com/decorator-factory/6b5276ab5fb87f4c009f1d22dc4b20a9

Sometimes it also just hangs, with 0 CPU consumption. When I press CTRL+C, I get this traceback: https://gist.github.com/decorator-factory/888a584cbc4d44600ea62f5e7e344b66

Sometimes it completes successfully.

graingert:
Fixes #414

the I couldn't get the original MRE from the issue https://gist.github.com/decorator-factory/6b5276ab5fb87f4c009f1d22dc4b20a9 to always fail - it seems to be intermittent. However the following *always* fails for me without this patch:

```python
import os
import asyncio
import subprocess

import uvloop
uvloop.install()


async def main():
    done = False
    def preexec_fn():
        pass

    async def fork():
        nonlocal done
        try:
            for i in range(1000):
                await asyncio.sleep(0)
                try:
                    proc = await asyncio.create_subprocess_exec("true", preexec_fn=preexec_fn)
                    await proc.communicate()
                except subprocess.SubprocessError:
                    print("done")
        finally:
            done = True

    def write_file():
        while not done:
            with open("demo", "w") as f:
                f.write("hello")
                print(f.fileno())

    await asyncio.gather(
        fork(),
        asyncio.to_thread(write_file)
    )

asyncio.run(main())
```

### DIFF
--- a/uvloop/handles/process.pyx
+++ b/uvloop/handles/process.pyx
@@ -101,6 +101,7 @@ cdef class UVProcess(UVHandle):
             self._finish_init()
 
             os_close(self._errpipe_write)
+            self._errpipe_write = -1
 
             if preexec_fn is not None:
                 errpipe_data = bytearray()


uvloop/handles/process.pyx
```diff
@@ -101,6 +101,7 @@ cdef class UVProcess(UVHandle):
             self._finish_init()
 
             os_close(self._errpipe_write)
+            self._errpipe_write = -1
```
graingert, line 104: the write_pipe can be closed again sometimes:

https://github.com/MagicStack/uvloop/blob/3e71ddc3383a8e0546519117c8775323c19eb6d7/uvloop/handles/process.pyx#L118-L122

if another thread allocates a file descriptor in this time then instead of raising OSError here the fd will be closed incorrectly

graingert:
@fantix can I get a review of this please?

fantix (approved):
This is good; I was trying to write a test but it's going to take some time, let's merge this first.

Issue #468, opened 2022-05-13
TimerHandle.when() segfaults if called after callback


j0x539:
* **uvloop version**: uvloop==0.16.0
* **Python version**: 3.10.4
* **Platform**: Linux-5.17.5-200.fc35.x86_64-x86_64-with-glibc2.34
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: Yes
* **Does uvloop behave differently from vanilla asyncio? How?**: It segfaults.

Calling when() on a TimerHandle object after callback has been executed causes a core dump. 

```python
import uvloop
import asyncio


async def main():
    loop = asyncio.get_running_loop()
    timer = loop.call_later(0.1, lambda: None) # call any function
    await asyncio.sleep(0.2)
    timer.when() # segfault happens here
    
uvloop.install()
asyncio.run(main())

fantix:
Thanks for the report and the PR! This is now fixed in c39afff.

fantix:
This is now included in 0.17.0.

PR #469, opened 2022-05-14
TimerHandle.when() segfault fix

j0x539:
Calling TimerHandle.when() after the timer has fired will cause a segmentation fault because when the timer completes self._clear() is called which will set timer to None.


graingert:
Fixes https://github.com/MagicStack/uvloop/issues/468

PR #470, opened 2022-05-15
cache when when creating TimerHandle

### LINKED ISSUE #468: TimerHandle.when() segfaults if called after callback

graingert:
fixes #468

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -857,9 +857,9 @@ class TestBaseUV(_TestBase, UVTestCase):
         delay = 1.0
         loop_t = self.loop.time()
         handle = self.loop.call_later(delay, cb)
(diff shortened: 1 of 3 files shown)


fantix:
Thanks for the PR! A similar approach is committed at c39afff.

PR #475, opened 2022-06-21
fix #469 segfault in timer.when() after it has fired/cancelled

jensbjorgensen:
instead cache that value at start-time

(and also)
ignore uvloop-dev (especially since dev instructions specify that you create it)
fix loop.pyi _TransProtPair which fails in python >=3.8


jensbjorgensen:
@elprans have a minute to take a look at this? it'd be good to fix this segfault... also if you have a chance to look at my feature pull request for the filesystem monitoring that would be super cool

fantix:
Thank you for the PR! I think it's overall good, though I'd do it slightly differently. Merged in c39afff and fae5f7f.

PR #477, opened 2022-07-13
Upgrade GitHub Actions

cclauss:
"3.10" must be quoted in yaml... https://dev.to/hugovk/the-python-3-1-problem-85g and #455


fantix (approved):
Thanks!

PR #480, opened 2022-07-17
Upgrade GitHub Actions for release process

cclauss:


.github/workflows/release.yml
```diff
@@ -50,19 +50,21 @@ jobs:
...
 
-    - uses: actions/setup-python@v2
+    - uses: actions/setup-python@v4
+      with:
+        python-version: 3.x
 
     - name: Build source distribution
       run: |
-        python -m pip install -U setuptools wheel pip
+        python -m pip install --upgrade setuptools wheel pip
```
cclauss, line 64: pip install has [command line options](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-user) for:
    --upgrade
    --upgrade-strategy
    --user
    --use-pep517

In a workflow, it is better to be explicit about which of these four we are using.

fantix (approved):

ben9923:
@fantix @cclauss I think those changes will make the release workflow fail. See https://github.com/MagicStack/uvloop/pull/447#issuecomment-1167953412.

PR #481, opened 2022-07-19
use a stack of self._fds_to_close to prevent double closes

graingert:
and make tests easier to write because the close order is deterministic
and in the order that opens happen in

this should also be a bit faster because `list.append` is faster
than `set.add` and we skip a call to `os_close(-1)` and catching an
`OSError` exception

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -7,6 +7,7 @@ import signal
 import subprocess
 import sys
 import tempfile
+import textwrap
 import time
 import unittest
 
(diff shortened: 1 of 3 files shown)


tests/test_process.py
```diff
@@ -796,7 +797,58 @@ async def test():
...
+                "os.pipe", new=log_pipes
+            ), mock.patch(
+                "os.dup", new=log_dups
+            ):
+                import uvloop
+
+
+            async def test():
+                proc = await asyncio.create_subprocess_exec(
+                    sys.executable, "-c", "pass"
```
graingert, line 834: you don't actually need a preexec_fn to trigger the double close it just made it slow enough that the race was more detectable

uvloop/handles/process.pyx
```diff
@@ -115,17 +117,8 @@ cdef class UVProcess(UVHandle):
...
         finally:
-            os_close(self._errpipe_read)
-            try:
-                os_close(self._errpipe_write)
-            except OSError:
-                # Might be already closed
-                pass
-
-            fds_to_close = self._fds_to_close
-            self._fds_to_close = None
```
fantix, line 126: I think `self._fds_to_close = None` is still needed?

graingert: Ah yep I see it

graingert: I moved it up so that _close_after_spawn can't interfere with the errpipe_write fd

uvloop/handles/process.pyx
```diff
@@ -69,6 +69,10 @@ cdef class UVProcess(UVHandle):
                 'Racing with another loop to spawn a process.')
 
         self._errpipe_read, self._errpipe_write = os_pipe()
+        fds_to_close = self._fds_to_close
+        self._fds_to_close = None
+        fds_to_close.append(self._errpipe_read)
+        fds_to_close.append(self._errpipe_write)
```
graingert, line 75:
```suggestion
        # add the write pipe last so we can close it early
        fds_to_close.append(self._errpipe_write)
```

uvloop/handles/process.pyx
```diff
@@ -500,10 +496,7 @@ cdef class UVProcessTransport(UVProcess):
                     # shouldn't ever happen
                     raise RuntimeError('cannot apply subprocess.STDOUT')
 
-                newfd = os_dup(io[1])
-                os_set_inheritable(newfd, True)
-                self._close_after_spawn(newfd)
-                io[2] = newfd
+                io[2] = self._file_redirect_stdio(io[1])
```
graingert, line 499: when reviewing all the calls to `_close_after_spawn` to make sure a set wasn't needed, I found this duplicate code 

https://github.com/MagicStack/uvloop/blob/5926386f81ea49903db9d5095058787ed91cff3d/uvloop/handles/process.pyx#L398-L402

graingert:
@fantix can I get a review on this please?

uvloop/handles/process.pxd
```diff
@@ -8,7 +8,7 @@ cdef class UVProcess(UVHandle):
         object _preexec_fn
         bint _restore_signals
 
-        set _fds_to_close
+        list _fds_to_close
```
graingert, line 11: @fantix I think there's a proper `list[int]` data type in cython for this but I don't know how to use it

fantix (approved):

graingert:
@fantix thanks for the review!

PR #484, opened 2022-08-02
enable the cython binding=True directive

graingert:
it becomes the default in cython3 and is needed for sniffio
to detect uvloop as asyncio efficiently

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -598,6 +598,16 @@ class _TestBase:
         self.loop.set_task_factory(None)
         self.assertIsNone(self.loop.get_task_factory())
 
+    def test_asyncgen_hooks(self):
+        async def get_asyncgens():
+            return sys.get_asyncgen_hooks()
+
+        # for sniffio to detect uvloop as asyncio efficiently the running loop
+        # call_soon module needs to match the asyncgen_hooks module
+        hooks = self.loop.run_until_complete(get_asyncgen_hooks)
+        self.assertEqual(hooks.finzalier.__module__, "uvloop.loop")
+        self.assertEqual(self.loop.call_soon.__module__, "uvloop.loop")
+
     def test_shutdown_asyncgens_01(self):
         finalized = list()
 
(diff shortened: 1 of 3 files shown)


tests/test_base.py
```diff
@@ -598,6 +598,16 @@ async def coro():
...
         self.assertIsNone(self.loop.get_task_factory())
 
+    def test_asyncgen_hooks(self):
+        async def get_asyncgens():
+            return sys.get_asyncgen_hooks()
+
+        # for sniffio to detect uvloop as asyncio efficiently the running loop
+        # call_soon module needs to match the asyncgen_hooks module
+        hooks = self.loop.run_until_complete(get_asyncgen_hooks)
+        self.assertEqual(hooks.finzalier.__module__, "uvloop.loop")
```
graingert, line 608:
```suggestion
        self.assertEqual(hooks.finalizer.__module__, "uvloop.loop")
```

tests/test_base.py
```diff
@@ -598,6 +598,16 @@ async def coro():
...
 
+    def test_asyncgen_hooks(self):
+        async def get_asyncgens():
+            return sys.get_asyncgen_hooks()
+
+        # for sniffio to detect uvloop as asyncio efficiently the running loop
+        # call_soon module needs to match the asyncgen_hooks module
+        hooks = self.loop.run_until_complete(get_asyncgen_hooks)
+        self.assertEqual(hooks.finalizer.__module__, "uvloop.loop")
+        self.assertEqual(self.loop.call_soon.__module__, "uvloop.loop")
```
graingert, line 609:
```suggestion
        async def acall(fn):
            return fn()

        # for sniffio to detect uvloop as asyncio efficiently the running loop
        # call_soon module needs to match the asyncgen_hooks module
        hooks = self.loop.run_until_complete(acall(sys.get_asyncgen_hooks))
        self.assertEqual(hooks.finalizer.__module__, "uvloop.loop")
        self.assertEqual(self.loop.call_soon.__module__, "uvloop.loop")
```

graingert:
Hmm, it might be worth updating to cython 3 instead?

fantix:
Oh let me try that

PR #491, opened 2022-09-09
Expose uv_version() for libuv API compatibility

fantix:
Refs #310, users may now use the `uvloop.loop.get_uv_version()` to check libuv API compatibility when integrating with `loop.get_uv_loop_t_ptr()`.


1st1:
Let's rename to `uvloop.loop.libuv_get_version()` and `uvloop.loop.libuv_get_loop_t()`

1st1:
Also the `uvloop.loop.libuv_get_version()` doesn't have to be a loop method, or does it?

fantix:
> Also the `uvloop.loop.libuv_get_version()` doesn't have to be a loop method, or does it?

`get_version()` doesn't have to be a loop method - it just reveals the version of the libuv library which uvloop is linking to.

1st1 (approved):

PR #495, opened 2022-09-11
Fix loop.getaddrinfo() and tests

### LINKED ISSUE #494: Inconsistent static getaddrinfo() behavior on AI_CANONNAME
Testing script:

```python
import socket
print(
    socket.getaddrinfo(
        '127.0.0.1',
        80,
        type=socket.SOCK_STREAM,
    )[0][3],  # ai_canonname
    socket.getaddrinfo(
        '127.0.0.1',
        80,
        type=socket.SOCK_STREAM,
        flags=socket.AI_CANONNAME,
    )[0][3],  # ai_canonname
)
```

Testing result:

| Platform    | flags=AI_CANONNAME | flags=0       |
|-------------|--------------------|---------------|
| Linux glibc | `"127.0.0.1"`      | `""`          |
| Linux musl  | `"127.0.0.1"`      | `"127.0.0.1"` |
| macOS       | `""`               | `""`          |
| uvloop      | `""`               | `""`          |

~Expected result: uvloop `__static_getaddrinfo_pyaddr()` returns consistent result as the platform std C lib returns.~

UPDATE: we would expect `__static_getaddrinfo_pyaddr()` honor the AI_CANONNAME flag.

fantix:
* ai_canonname is different across libc impls (#494)
* `AddressFamily` and `SocketKind` can be enums (copied from [cpython](https://github.com/python/cpython/blob/a36235d5c7863a85fa323b2048d3d254116a958e/Lib/socket.py#L100-L108))
* Also fixed failing test

Closes #494 

Per https://www.openwall.com/lists/musl/2013/03/29/13:

> it's a bug to assume a certain (libc) implementation has particular properties rather than testing.

~So the idea here is to run the libc/libuv `getaddrinfo()` first when we need to run `__static_getaddrinfo_pyaddr()` for the first time with both `AI_CANONNAME` on and off each, and remember if `ai_canonname` is set or not so that we can do the same in the future calls to `__static_getaddrinfo_pyaddr()`.~

Update: uvloop static resolving always honor the AI_CANONNAME flag, while libuv-based resolving is still consistent with the libc impl.

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1523,13 +1523,31 @@ cdef class Loop:
     @cython.iterable_coroutine
     async def getaddrinfo(self, object host, object port, *,
                           int family=0, int type=0, int proto=0, int flags=0):
+        global __static_getaddrinfo_canonname_mode
 
         addr = __static_getaddrinfo_pyaddr(host, port, family,
                                            type, proto, flags)
         if addr is not None:
-            fut = self._new_future()
-            fut.set_result([addr])
-            return await fut
+            if __static_getaddrinfo_canonname_mode & (
+                1 << 0 if flags & socket_AI_CANONNAME else 1 << 2
+            ):
+                return [addr]
+
+            rv = await self._getaddrinfo(
+                host, port, family, type, proto, flags, 1)
+
+            if flags & socket_AI_CANONNAME:
+                if rv[0][3]:
+                    __static_getaddrinfo_canonname_mode |= 1 << 0 | 1 << 1
+                else:
+                    __static_getaddrinfo_canonname_mode |= 1 << 0
+            else:
+                if rv[0][3]:
+                    __static_getaddrinfo_canonname_mode |= 1 << 2 | 1 << 3
+                else:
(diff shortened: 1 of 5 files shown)


uvloop/loop.pyx
```diff
@@ -1523,13 +1523,31 @@ cdef class Loop:
...
+            if __static_getaddrinfo_canonname_mode & (
+                1 << 0 if flags & socket_AI_CANONNAME else 1 << 2
+            ):
+                return [addr]
+
+            rv = await self._getaddrinfo(
+                host, port, family, type, proto, flags, 1)
+
+            if flags & socket_AI_CANONNAME:
+                if rv[0][3]:
```
1st1, line 1540: Please use proper constants

1st1: With good names

fantix: Fixed!

1st1 (approved):

fantix:
As Elvis suggested, we should just follow the docs, regardless of how different libc impls are behaving, and tweak the tests to pass.

PR #498, opened 2022-09-13
bugfix: write to another transport in resume_writing() fails

### LINKED ISSUE #496: RuntimeError: set changed size during iteration
* **uvloop version**:  0.16.0
* **Python version**: 3.10.2
* **Platform**: 64-bit Linux
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: not able to test
* **Does uvloop behave differently from vanilla asyncio? How?**: not able to test

I run a fastapi application with uvicorn, and I occassionally receive the following error:

```
  File "uvloop/cbhandles.pyx", line 66, in uvloop.loop.Handle._run
  File "uvloop/loop.pyx", line 473, in uvloop.loop.Loop._on_idle
  File "uvloop/loop.pyx", line 649, in uvloop.loop.Loop._exec_queued_writes
RuntimeError: set changed size during iteration
```

This seems to only happen in a production environment and I don't have the resources to examine it further. However, it seems that the code in the uvloop master branch is different, so maybe the error has been fixed already.

fantix:
In `resume_writing()`, writing to another transport may call `loop._queue_write()` hence mutating `loop._queued_streams`, it'll be an error if `resume_writing()` was called from within iterating `loop._queued_streams` in `loop._exec_queued_writes()`.

Fixes #496

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -645,25 +645,17 @@ cdef class Loop:
 
         cdef:
             UVStream stream
-            int queued_len
 
-        if UVLOOP_DEBUG:
-            queued_len = len(self._queued_streams)
+        streams = list(self._queued_streams)
+        self._queued_streams.clear()
 
-        for pystream in self._queued_streams:
+        for pystream in streams:
             stream = <UVStream>pystream
             stream._exec_write()
 
-        if UVLOOP_DEBUG:
-            if len(self._queued_streams) != queued_len:
-                raise RuntimeError(
-                    'loop._queued_streams are not empty after '
-                    '_exec_queued_writes')
-
-        self._queued_streams.clear()
-
         if self.handler_check__exec_writes.running:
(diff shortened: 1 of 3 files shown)


uvloop/loop.pyx
```diff
@@ -645,25 +645,17 @@ cdef class Loop:
 
         cdef:
             UVStream stream
-            int queued_len
 
-        if UVLOOP_DEBUG:
-            queued_len = len(self._queued_streams)
+        streams = list(self._queued_streams)
```
1st1, line 649: Can we have 2 python `list` objects that are always allocated and switch them here?

```
streams = self._queued_streams
self._queued_streams = self._queued_streams2
try:
   ...
finally:
   streams.clear()
```

fantix: Sure, only that it'll be 2 `set`s.

1st1 (approved):

Issue #471, opened 2022-05-21
aiohttp-client connections cannot be really closed when uvloop installed


Rongronggg9:
* **uvloop version**: 0.16.0
* **Python version**: CPython 3.9.12
* **Platform**: Debian GNU/Linux bookworm/sid x86_64
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: yes
* **Does uvloop behave differently from vanilla asyncio? How?**: yes, the bug is only reproducible with uvloop

# Desciption

If `uvloop` is installed, even if the response, connection, and session are all closed, the server from which aiohttp requested resources, will still keep sending packets to the aiohttp-client for some time. However, the bug is only reproducible with uvloop installed.

# To reproduce

```
aiohttp[speedups]==3.8.1
uvloop==0.16.0
```

```py
import asyncio
import aiohttp
import uvloop
import logging

logging.basicConfig(level=logging.DEBUG)


async def fetch(url):
    async with aiohttp.ClientSession() as session:
        async with session.get(url) as response:
            logging.debug(response)
            logging.debug(response.content)


async def reproduce(wait):
    await fetch('https://sgp-ping.vultr.com/vultr.com.1000MB.bin')
    await asyncio.sleep(wait)  # now you will see that the file is still being downloaded


def main(enable_uvloop):
    if enable_uvloop:
        logging.info('********** Using uvloop **********')
        uvloop.install()  # the bug is only reproducible with uvloop
    else:
        logging.info('********** Using asyncio **********')
    loop = asyncio.new_event_loop()
    loop.set_debug(True)  # or set env PYTHONASYNCIODEBUG=1, not necessary to reproduce the bug
    asyncio.set_event_loop(loop)
    loop.run_until_complete(reproduce(15 if enable_uvloop else 5))
    loop.close()


if __name__ == '__main__':
    main(enable_uvloop=False)
    main(enable_uvloop=True)
    input('Press Enter to exit')
```

# Log and screenshot
```
INFO:root:********** Using asyncio **********
DEBUG:asyncio:Using selector: EpollSelector
DEBUG:asyncio:Get address info sgp-ping.vultr.com:443, type=<SocketKind.SOCK_STREAM: 1>, flags=<AddressInfo.AI_ADDRCONFIG: 32>
[...]

PR #507, opened 2022-10-01
Close transport after sending close_notify in TLSv1.2

fantix:
See #471 for details.


stalkerg:
What is the problem with merging it? Seem like we have huge problems with aiohttp + AWS services

PR #511, opened 2022-11-01
 feat(uvloop): Support event loop utilization queries

caseydialpad:
This change seeks to expose `libuv`'s internal [`uv_metrics_idle_time`](https://docs.libuv.org/en/v1.x/metrics.html#c.uv_metrics_idle_time) to permit users to tune their applications. This change is inspired by Nodejs' [`internalEventLoopUtilization`](https://github.com/nodejs/node/blob/v19.0.0/lib/internal/perf/event_loop_utilization.js#L22-L68).

The heart of this change is:

```python

 def event_loop_utilization(self):
    """Returns idle and active time in milliseconds and the percentage of
    time the event loop is active
    """

    ...

    return idle, active, utilization
```


caseydialpad:
Tests will be fixed by a change like https://github.com/MagicStack/uvloop/pull/512.

PR #519, opened 2022-12-17
Update CI tests to use released 3.11

altendky:
Draft for:
- [ ] Fix general CI, such as the flake8 failure https://github.com/MagicStack/uvloop/actions/runs/3720859641/jobs/6310624775#step:7:570

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-rc.1"]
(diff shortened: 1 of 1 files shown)


altendky:
I'll note that you can set up the workflow to not require adding and removing this if you would like.  Just add, for example, `3.12` to the list when you want to start testing it and track from alpha through released without changes.  You wouldn't need the pypy aspects but https://github.com/pytest-dev/pytest-twisted/blob/56991132000791cf89283ac5aa033f3469c63e35/.github/workflows/ci.yml#L154-L167 shows the gist of it.  Happy to offer a PR if you are interested.

fantix:
Thanks for the PR! Closing as #553 is merged.

Re: ranged version for setup-python action: sure, why not ;)

PR #523, opened 2023-01-11
fix github workflow badge

EFord36:
At the moment, the badges show like:

Screenshot 2023-01-11 at 13 25 12

See the linked GitHub issue from the previous version of the badge for the root cause: https://github.com/badges/shields/issues/8671

In summary, this type of badge was changed in a breaking fashion such that the badges were just linking to the github issue rather than showing the data. Updating the url resolves this.

Also update the link of the badge to specify the workflow file rather than the name, which matches the new behaviour of the badge.

Let me know if I need to do the Sphinx build - it looks to me like this is done within ReadTheDocs but let me know if there's anything I need to do here.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
-.. image:: https://img.shields.io/github/workflow/status/MagicStack/uvloop/Tests
-    :target: https://github.com/MagicStack/uvloop/actions?query=workflow%3ATests+branch%3Amaster
+.. image:: https://img.shields.io/github/actions/workflow/status/MagicStack/uvloop/tests.yml?branch=master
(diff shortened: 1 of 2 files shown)


EFord36:
Ah, I see that my PR is basically a duplicate of the earlier #522 - sorry, I didn't notice that when opening it somehow, I thought I had checked open PRs and issues :/

The only difference is that my PR also changes the link targets to link to the specific file, which matches the behaviour of the badge more exactly. I think this is very minor though, and that the author of #522 should get the credit here since they beat me to it!

Happy to close, leaving open for now due to the slight difference.

fantix:
No problem! I'll merge both in one squashed commit. Thanks for the PR!

fantix:
Merged in #522, thanks for the PR!

PR #535, opened 2023-04-07
Add thread timeout for loop.shutdown_default_executor

Gelbpunkt:
Ref: https://github.com/python/cpython/commit/575a253b5c203e8d2ebfd239ed5a613179f8984f


fantix (approved):
Thanks for the PR! As we got most of the same changes in #570, I'll accept the docstring and `warnings_warn` shortcut fix.

PR #545, opened 2023-06-13
if send(to) fails with OSError should result in protocol.error_received

jensbjorgensen:
I have observed a case where due to networking change an attempt at sending (sendto) resulted in our code receiving connection_lost on the protocol object with exc=OSError(101, 'Network is unreachable'). This is inconsistent with the upstream CPython asyncio implementation, which will call error_received on the protocol object in the event of isinstance(exc, OSError). 

This is fairly serious because it results in the socket being closed whereas in the core asyncio implementation further send(to)'s would succeed when the network condition was rectified.

### DIFF
--- a/uvloop/_version.py
+++ b/uvloop/_version.py
@@ -10,4 +10,4 @@
 # supported platforms, publish the packages on PyPI, merge the PR
 # to the target branch, create a Git tag pointing to the commit.
 
-__version__ = '0.18.0'
+__version__ = '0.19.0'


jensbjorgensen:
Interestingly, in the uv_udp_send_cb callback case OSError errors are treated that way but an immediate error return is also possible.

jensbjorgensen:
I don't think the failed tests are due to a problem with my changes

PR #554, opened 2023-08-11
bump libuv to 1.46.0

dulmandakh:
there are many bug fixes and improvements, like io_uring support https://www.phoronix.com/news/libuv-io-uring

### DIFF
--- a/tests/test_process.py
+++ b/tests/test_process.py
@@ -115,7 +115,7 @@ class _TestProcess:
                 cmd, b'-W', b'ignore', '-c',
                 'import os,sys;sys.stdout.write(os.getenv("FRUIT"))',
                 stdout=subprocess.PIPE,
-                preexec_fn=lambda: os.putenv("FRUIT", "apple"))
+                env=dict(os.environ, FRUIT="apple"))
 
             out, _ = await proc.communicate()
             self.assertEqual(out, b'apple')


achimnol:
Would this enable writing true async file I/O implementation aside the current thread-pool based solutions such as aiofiles? 👀

dulmandakh:
> Would this enable writing true async file I/O implementation aside the current thread-pool based solutions such as aiofiles? 👀
> Since there is no official `AbstractEventLoop` interface to offer async file I/O yet, I think uvloop may need to define additional methods to expose the underlying libuv's file I/O APIs and write a high-level wrapper as a separate module or package.

I hope that it'll make networking much faster for Python.

achimnol:
> > Would this enable writing true async file I/O implementation aside the current thread-pool based solutions such as aiofiles? 👀
> > Since there is no official `AbstractEventLoop` interface to offer async file I/O yet, I think uvloop may need to define additional methods to expose the underlying libuv's file I/O APIs and write a high-level wrapper as a separate module or package.
> 
> I expect that it'll make networking much faster for Python.

That will be also the effect, but what I'm looking forward to see is the native async operations on disk file I/O without using thread pools.

fantix:
Thank you for the PR! I'm closing this one as #571 is merged. Regarding io_uring-based file I/O, another Python package as a uvloop extension is probably the most feasible way to implement it - I'm open to adding libuv-specific Cython APIs to support extensions.

PR #534, opened 2023-03-18
Mark functions as noexcept to ensure cython 3 compatibility

matusvalo:
Cython 3 will require `noexcept` keyword for functions not raising an exception: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept
Cython 0.29.32 introduces this keyword too for forward compatibility. This PR fixes multiple issues found in https://github.com/cython/cython/issues/5305.

> **Note**: Currently uvloop cannot be compiled in cython 3.0.0b1 because of issue https://github.com/cython/cython/issues/5270. In the next Cython release the issue should be fixed.

PR #555, opened 2023-08-12
CI: python 3.12-dev

dulmandakh:
This PR is blocked because _imp_ module used by Cython is removed in Python 3.12, thus causing import errors. Cython switched to _importlib_ in version 3.0.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
         os: [ubuntu-latest, macos-latest]
 
     env:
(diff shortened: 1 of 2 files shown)


michael-k:
Is there a way forward/a way to unblock this?  Is MagicStack/uvloop#534 enough to unblock this?

tarsil:
By the time of this writing, 3.0.3 was merged and a release is being prepared https://github.com/cython/cython/actions/runs/6417571100

Maybe we can finally move on :)

.github/workflows/tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
```
josecsotomorales, line 17: We should use `3.12` instead of `3.12-dev`

josecsotomorales (changes requested):

tarsil:
This PR is in draft mode. Maybe someone needs to create another one with the official move to python 3.12 and also take care of this 

https://github.com/MagicStack/uvloop/issues/568

fantix:
Yeah, thank you for the PR! Closing because #570 is merged.

PR #559, opened 2023-09-04
fix deadlock in test_remote_shutdown_receives_trailing_data

xiacunshun:
The hanging is due to `eof_received = threading.Lock()` hold by client which is requested by server.
The client keep the lock because the `reader.read()` does not return as expected.
So I try to use `asyncio.wait_for` to make it return if nothing recieved.


1st1:
@fantix please take a look

xiacunshun:
Is there anyboby can deal with this?

fantix:
Yeah, sorry for the delay! I'm looking into this.

xiacunshun:
@fantix Is there any problem?

xiacunshun:
@fantix

PR #570, opened 2023-10-08
Port uvloop to Python 3.12

### LINKED ISSUE #569: Can we update cython and make uvloop compatible with python 3.12 please?
* **uvloop version**: 0.17.0
* **Python version**: 3.12
* **Platform**: all
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: N/A
* **Does uvloop behave differently from vanilla asyncio? How?**: N/A

Cython was recently updated and this allows uvloop to be compatible with python 3.12. There are a plethora of projects out there that relies on this. 

Is it possible to prioritize this and make a release, please? Asyncpg, for instance, it's one of them. 

Appreciate the understanding,
tarsil

1st1:
Fixes #569


1st1:
While working on this I stumbled upon what seems to be an asyncio regression in 3.12. Good news that uvloop doesn't have this regression. Bad news -- our tests that test compliance with asyncio are now failing (we should backport our functional tests to the Python project, they increase test coverage by a lot!)

setup.py
```diff
@@ -37,6 +36,15 @@
     CYTHON_DEPENDENCY,
 ]
 
+if vi < (3, 12):
+    # XXX Revert this change later.
+    #
+    # Python 3.12 is new and there's no aiohttp wheel for it yet.
+    # And pip helfully fails on building a wheel for it and I've
+    # no idea how to disable that.
```
graingert, line 44: There's an environment variable you can set to use the pure Python http parser

README.rst
```diff
@@ -53,6 +53,27 @@ uvloop with::
 Using uvloop
 ------------
 
+Starting with Python 3.12, the preferred way of using uvloop
```
graingert, line 56: uvloop.run seems to be available on all supported Python versions so I think this documentation should be "as of uvloop x.y, the preferred way of..."

graingert (changes requested):

README.rst
1st1: good one, thanks

setup.py
1st1: The problem isn't using aiohttp, it's about not being able to install it with 3.12 in CI

graingert: I know, the environment variable will let you install it in CI

graingert: Add this to your workflow and the aiohttp wheel will build and install from pypi

```
      - name: Set AIOHTTP_NO_EXTENSIONS=1 on 3.12
        if: matrix.environment == '3.12'
        run: echo AIOHTTP_NO_EXTENSIONS=1 >> $GITHUB_ENV
        shell: bash
```

fantix: Hmm, actually, I'll just use `aiohttp== 3.9.0b0`, which was only released 2 days ago.

fantix (approved):

AntonOfTheWoods:
Is there an ETA on a release?

PR #573, opened 2023-10-14
restore uvloop.new_event_loop and other missing uvloop members to typing

graingert:


uvloop/__init__.pyi
```diff
@@ -5,10 +5,25 @@ from collections.abc import Callable, Coroutine
...
 
-
 _T = TypeVar('_T')
 
 
+class Loop(AbstractEventLoop):
+    ...
+
+
+def new_event_loop() -> Loop:
```
graingert, line 15: this is needed for `async with asyncio.Runner(loop_factory=uvloop.new_event_loop)`

uvloop/__init__.pyi
graingert, line 1: adding this file removed the useful members like 'new_event_loop' which is needed for `asyncio.Runner(...)`

uvloop/__init__.py
```diff
@@ -9,10 +11,12 @@
 from .loop import Loop as __BaseLoop  # NOQA
 from ._version import __version__  # NOQA
 
-
 __all__ = ('new_event_loop', 'install', 'EventLoopPolicy')
```
graingert, line 14:
```suggestion

__all__ = ('new_event_loop', 'install', 'EventLoopPolicy')
```

uvloop/__init__.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
graingert, line 2:
```suggestion
```

uvloop/__init__.py
```diff
@@ -13,6 +13,9 @@
 __all__ = ('new_event_loop', 'install', 'EventLoopPolicy')
 
 
+_T = TypeVar("_T")
```
graingert, line 16:
```suggestion
_T = _typing.TypeVar("_T")
```

uvloop/__init__.py
```diff
@@ -1,6 +1,6 @@
 import asyncio as __asyncio
 import typing as _typing
-import sys as _sys
```
graingert, line 3: mypy doesn't recognise `_sys.version_info` in comparisons it needs the real `sys.version_info`

uvloop/__init__.py
```diff
@@ -34,10 +37,16 @@ def install() -> None:
     __asyncio.set_event_loop_policy(EventLoopPolicy())
 
 
-def run(main, *, loop_factory=new_event_loop, debug=None, **run_kwargs):
+def run(
+    main: _typing.Coroutine[_typing.Any, _typing.Any, _T],
+    *,
+    loop_factory: _typing.Optional[_typing.Callable[[], Loop]] = new_event_loop,
+    debug: _typing.Optional[bool]=None,
+    **run_kwargs: object
```
graingert, line 45: this is a bit problematic with mypy typing, it's a lie that run_kwargs exists but it's needed for backwards compatibility

graingert: might be best to do this whole thing in an:

```
if _typing.TYPE_CHECKING:
    def run(...): ...
else:
    def run(..., **run_kwargs): ...
```

graingert:
@fantix can I get a review on this please? It's blocking mypy in unicorn

uvloop/__init__.py
```diff
@@ -34,69 +37,84 @@ def install() -> None:
...
-            raise ValueError("a coroutine was expected, got {!r}".format(main))
-
-        loop = loop_factory()
-        try:
-            __asyncio.set_event_loop(loop)
-            if debug is not None:
-                loop.set_debug(debug)
-            return loop.run_until_complete(wrapper())
-        finally:
+if _typing.TYPE_CHECKING:
```
fantix, line 40: Can we just inline the typing and omit this TYPE_CHECKING check?

graingert: I tried that, if you see the history of the PR, but it took too many changes

fantix: I see now. What about adding `new_event_loop` to `__init__.pyi`?

fantix (approved):

uvloop/__init__.py
graingert: I tried that first also and it means the other members aren't type checked

PR #590, opened 2024-01-17
fix Future exception never retrieved in create_connection

jensbjorgensen:
I also have observed that when I use asyncio.wait_for(...) around loop.create_connection in the timeout case I end up getting this logged to stdout(err?):

Future exception was never retrieved
future: <Future finished exception=gaierror(-2, 'Name or service not known')>
socket.gaierror: [Errno -2] Name or service not known

The code in this commit fixes this.

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1972,7 +1972,12 @@ cdef class Loop:
                     lai = &lai_static
 
             if len(fs):
-                await aio_wait(fs)
+                try:
+                    await aio_wait(fs)
+                except asyncio.CancelledError as exc:
+                    for fut in fs:
+                        fut.cancel()
+                    raise exc from None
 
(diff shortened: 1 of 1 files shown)


uvloop/loop.pyx
```diff
@@ -1972,7 +1972,12 @@ cdef class Loop:
                     lai = &lai_static
 
             if len(fs):
-                await aio_wait(fs)
+                try:
+                    await aio_wait(fs)
+                except asyncio.CancelledError as exc:
+                    for fut in fs:
+                        fut.cancel()
+                    raise exc from None
```
fantix, line 1980: Why dropping the original context here, instead of just `raise`?

fantix (commented):
This makes sense! Though a unit test is preferred.

PR #591, opened 2024-01-18
Use len(os.sched_getaffinity(0)) instead of os.cpu_count()

avkarenow:
This change allows to control number of processes by cpuset command.


fantix (approved):
Awesome! TIL

PR #608, opened 2024-06-18
really drop python<=3.7 support

kloczek:
Filter all code over `pyupgrade --py38-plus`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ MACHINE = platform.machine()
 MODULES_CFLAGS = [os.getenv('UVLOOP_OPT_CFLAGS', '-O2')]
 _ROOT = pathlib.Path(__file__).parent
 LIBUV_DIR = str(_ROOT / 'vendor' / 'libuv')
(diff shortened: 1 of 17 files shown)


setup.py
```diff
@@ -1,9 +1,5 @@
 import sys
 
-vi = sys.version_info
-if vi < (3, 8):
-    raise RuntimeError('uvloop requires Python 3.8 or greater')
-
```
fantix, line 6: > "requires-python = '>=3.8.0'" in pyproject.toml is enough.

I'd keep this extra check. For people who uses Python < 3.8, they may also have a `pip` around year 2016 or earlier, which doesn't honor `requires-python`.

tests/test_pipes.py
fantix, line 1: unused import `io` now

fantix (approved):
Thanks for the PR! I feel neutral about most of the changes here, but still approving.

stankudrow:
Any progress? Should I pick up this PR?

PR #609, opened 2024-06-20
Use cythonized SO_REUSEPORT rather than the unwrapped native one.

### LINKED ISSUE #550: uvloop strictly needs SO_REUSEPORT
uvloop expects that `SO_REUSEPORT` is supported everywhere.  Unfortunately, there are some platforms where `SO_REUSEPORT` is not supported.  For example illumos.

Error message:
```
uvloop/loop.c: In function '__pyx_gb_6uvloop_4loop_4Loop_57generator3':
uvloop/loop.c:33764:46: error: 'SO_REUSEPORT' undeclared (first use in this function); did you mean 'SO_REUSEADDR'?
33764 |             __pyx_t_5 = __Pyx_PyInt_From_int(SO_REUSEPORT); if (unlikely(!__pyx_t_5)) __PYX_ERR(0, 1778, __pyx_L32_error)
      |                                              ^~~~~~~~~~~~
      |                                              SO_REUSEADDR
uvloop/loop.c:33764:46: note: each undeclared identifier is reported only once for each function it appears in
```

* **uvloop version**: 0.17.0
* **Python version**: 3.9.16
* **Platform**: OpenIndiana/illumos
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: N/A
* **Does uvloop behave differently from vanilla asyncio? How?**: N/A

Could you please detect the `SO_REUSEPORT` support during build and do not try to use `SO_REUSEPORT` if it is not available?

Thank you.

ptribble:
The code already detects (via hasattr) whether SO_REUSEPORT is available, setting has_SO_REUSEPORT, and creates a wrapped SO_REUSEPORT (via getattr). This change ensures it's that wrapped SO_REUSEPORT that's used, rather than the native C version (which may not exist).

Fixes #550

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1775,7 +1775,7 @@ cdef class Loop:
                         if reuse_address:
                             sock.setsockopt(uv.SOL_SOCKET, uv.SO_REUSEADDR, 1)
                         if reuse_port:
-                            sock.setsockopt(uv.SOL_SOCKET, uv.SO_REUSEPORT, 1)
+                            sock.setsockopt(uv.SOL_SOCKET, SO_REUSEPORT, 1)
                         # Disable IPv4/IPv6 dual stack support (enabled by
                         # default on Linux) which makes a single socket
                         # listen on both address families.
--- a/uvloop/includes/uv.pxd
+++ b/uvloop/includes/uv.pxd
@@ -57,7 +57,7 @@ cdef extern from "uv.h" nogil:
     cdef int SOL_SOCKET
     cdef int SO_ERROR
     cdef int SO_REUSEADDR
-    cdef int SO_REUSEPORT
+    # use has_SO_REUSEPORT and SO_REUSEPORT in stdlib.pxi instead
     cdef int AF_INET
     cdef int AF_INET6
     cdef int AF_UNIX


uvloop/loop.pyx
```diff
@@ -1775,7 +1775,7 @@ cdef class Loop:
                         if reuse_address:
                             sock.setsockopt(uv.SOL_SOCKET, uv.SO_REUSEADDR, 1)
```
mtelka, line 1776: Isn't the same problem here with `SO_REUSEADDR`?

ptribble: No; the place to look is `uvloop/includes/stdlib.pxi`

    cdef int has_IPV6_V6ONLY = hasattr(socket, 'IPV6_V6ONLY')
    cdef int IPV6_V6ONLY = getattr(socket, 'IPV6_V6ONLY', -1)
    cdef int has_SO_REUSEPORT = hasattr(socket, 'SO_REUSEPORT')
    cdef int SO_REUSEPORT = getattr(socket, 'SO_REUSEPORT', 0)

Note that there's a check to see if the variable exists, and then a python definition of the variable with a default, guaranteeing it's defined. This is how the code handles platforms that don't have the constants defined. For other constants (such as `SO_REUSEADDR`) there's neither a check nor a python definition of the constant.

fantix (approved):
Let's also drop the definition of `uv.SO_REUSEPORT` to prevent future usage.

PR #614, opened 2024-08-11
Fix test_create_server_4 with Python 3.12.5

shadchin:
After https://github.com/python/cpython/issues/121913 error message `[errno 98] address already in use`


fantix (approved):
Thanks!

Issue #615, opened 2024-08-13
vulnerability in libuv v1.46.0


javaccar:
* **uvloop version**: 0.19.0
* **Python version**: 3.11
* **Platform**: linux
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: n/a
* **Does uvloop behave differently from vanilla asyncio? How?**: n/a

uvloop uses libuv v1.46.0, which has a security vulnerability https://nvd.nist.gov/vuln/detail/CVE-2024-24806
the vulnerability was fixed in libuv v1.48.0 but uvloop is still using v1.46.0.

fantix:
#600 would fix this

javaccar:
Thanks @fantix ! Appreciate the quick response. I see https://github.com/MagicStack/uvloop/pull/600 was just merged. Plans to cut a release and publish to pypi?

fantix:
No problem! Yes, I'll cut it tomorrow.

PR #604, opened 2024-04-30
Inline _Py_RestoreSignals() from CPython

befeleme:
private _Py_RestoreSignals() has been moved to CPython internals as of Python 3.13 See: https://github.com/python/cpython/pull/106400 Its implementation has been the same in all supported by uvloop Pythons (3.8+), so the inlining was not conditionalized.

Closes #603 

This allows to build uvloop with Python 3.13.0a6.

Unfortunately, one of the tests fails and I can't tell whether it's related to this or it's an incidental breakage coming from some other parts of CPython changes:

```
=================================== FAILURES ===================================
___________________ Test_AIO_Unix.test_create_unix_server_1 ____________________
Traceback (most recent call last):
  File "/usr/lib64/python3.13/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib64/python3.13/unittest/case.py", line 651, in run
    self._callTestMethod(testMethod)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/lib64/python3.13/unittest/case.py", line 606, in _callTestMethod
    if method() is not None:
       ~~~~~~^^
  File "/builddir/build/BUILD/uvloop-0.19.0/_empty/tests/test_unix.py", line 147, in test_create_unix_server_1
    self.loop.run_until_complete(start_server())
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 721, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/builddir/build/BUILD/uvloop-0.19.0/_empty/tests/test_unix.py", line 100, in start_server
    self.assertTrue(os.path.exists(sock_name))
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/unittest/case.py", line 744, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true
```

Issue #603, opened 2024-03-27
uvloop fails to build with Python 3.13.0a5: error: implicit declaration of function ‘_Py_RestoreSignals’


befeleme:
* **uvloop version**: 0.19.0
* **Python version**: 3.13.0a5
* **Platform**: Fedora Linux 41
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: n/a
* **Does uvloop behave differently from vanilla asyncio? How?**: n/a

I try to build uvloop for Fedora Linux 41 with Python 3.13.0a5.
The build fails with the information below.

```
  uvloop/loop.c: In function ‘__pyx_f_6uvloop_4loop_9UVProcess__after_fork’:
  uvloop/loop.c:129617:5: error: implicit declaration of function ‘_Py_RestoreSignals’ [-Werror=implicit-function-declaration]
  129617 |     _Py_RestoreSignals();
         |     ^~~~~~~~~~~~~~~~~~
```

befeleme:
`_Py_RestoreSignals()` has apparently been moved to internal CPython: https://github.com/python/cpython/commit/c9ce983ae1a361f431a0303aeb6f4b8e1d674275

hroncok:
https://github.com/python/cpython/blob/57aee2a02ce38be30cf6c310245547cf56562ab6/Modules/signalmodule.c#L1931-L1951

Maybe this can be inlined?

wRAR:
This one is a warning here on a pyenv-installed 3.13.0b1 (not sure if `-Werror=implicit-function-declaration` in the original report comes from Fedora defaults or what), but there are multiple actual errors:

```
      uvloop/loop.c:4446:105: error: invalid type argument of ‘->’ (have ‘int’)
       4446 |     __pyx_assertions_enabled_flag = ! _PyInterpreterState_GetConfig(__Pyx_PyThreadState_Current->interp)->optimization_level;
            |                                                                                                         ^~
[...]
      uvloop/loop.c: In function ‘__Pyx_PyInt_As_int’:
      uvloop/loop.c:177390:27: error: too few arguments to function ‘_PyLong_AsByteArray’
      177390 |                 int ret = _PyLong_AsByteArray((PyLongObject *)v,
             |                           ^~~~~~~~~~~~~~~~~~~
[...]
```

root-kidik:
+1

PR #600, opened 2024-03-12
Upgrade libuv to v1.48.0

niklasr22:
Upgrades libuv to [v1.48.0](https://github.com/libuv/libuv/releases/tag/v1.48.0) which fixes a security vulnerability.

I removed two DNS test cases because they raise an error intended by libuv.

PR #617, opened 2024-08-15
uvloop 0.20.0

fantix:
Changes
=======

* Upgrade libuv to v1.48.0 (#600)
  (by @niklasr22 @fantix in 77778525 for #596 #615)

Fixes
=====

* Fix test_create_server_4 with Python 3.12.5 (#614)
  (by @shadchin in 62f92393)

* Use len(os.sched_getaffinity(0)) instead of os.cpu_count() (#591)
  (by @avkarenow in c8531c24 for #591)

* Inline _Py_RestoreSignals() from CPython (#604)
  (by @befeleme in 8511ba1f for #603)

### DIFF
--- a/uvloop/_version.py
+++ b/uvloop/_version.py
@@ -10,4 +10,4 @@
 # supported platforms, publish the packages on PyPI, merge the PR
 # to the target branch, create a Git tag pointing to the commit.
 
-__version__ = '0.19.0'
+__version__ = '0.20.0'


edgarrmondragon:
Should 0.20.0 also include Python 3.13 wheels?

fantix:
> Should 0.20.0 also include (normal non-free-threaded) Python 3.13 wheels?
> 
> EDIT: Ok, too late for that 😅

Maybe 0.21 after your #610

PR #619, opened 2024-08-16
Improve performance and reduce latency of Transport.write by attempting to send data immediately if all write buffers are empty

tarasko:
Current implementation of UVStream.write almost never sends data immediately. Instead the data is stored in the buffer and picked up later by uv_check callback.
This introduces unnecessary latency and CPU overhead

This PR increases RPS rate between echoclient and echoserver by roughly 10%.
`echoclient --worker 1 --num 200000`
`echoserver --uvloop --proto `

Because  of this change a couple of test had to be tweaked.
* get_write_buffer_size cannot reliably return how much data has been already buffered. It doesn't take into account the socket send buffer (which is in the kernel)
* writing above limits does not necessarily cause pause_writing because again the data first goes to the socket send buffer

Please note that the current implementation of asyncio does the same thing. It tries to write directly to the socket, and, only if EWOULDBLOCK happens, the data is added to the buffer
https://github.com/python/cpython/blob/c13e7d98fb8581014a225b900b1b88ccbfc28097/Lib/asyncio/selector_events.py#L1065

### DIFF
--- a/uvloop/handles/stream.pyx
+++ b/uvloop/handles/stream.pyx
@@ -701,7 +701,7 @@ cdef class UVStream(UVBaseTransport):
             bytes_written = bytes_written_
 
             if bytes_written == 0:
-                # All data was successfully bytes_written.
+                # All data was successfully written.
                 # on_write will call "maybe_resume_protocol".
                 return
 
--- a/uvloop/handles/stream.pyx
+++ b/uvloop/handles/stream.pyx
@@ -424,11 +424,13 @@ cdef class UVStream(UVBaseTransport):
         self._buffer_size += dlen
         self._buffer.append(data)
 
-    cdef inline _initiate_write(self):
-        if (not self._protocol_paused and
-                (<uv.uv_stream_t*>self._handle).write_queue_size == 0 and
-                self._buffer_size > self._high_water):
+    cdef inline _initiate_write(self, bint skip_fast_path):
+        if (not skip_fast_path and
+            not self._protocol_paused and
+            (<uv.uv_stream_t*>self._handle).write_queue_size == 0 and
+            self._buffer_size > self._high_water):
             # Fast-path.  If:
+            #   - the caller hasn't tried fast path itself
             #   - the protocol isn't yet paused,
             #   - there is no data in libuv buffers for this stream,
             #   - the protocol will be paused if we continue to buffer data
@@ -687,9 +689,7 @@ cdef class UVStream(UVBaseTransport):
 
         cdef ssize_t bytes_written
 
-        if (not self._protocol_paused and self._buffer_size == 0 and
-            (<uv.uv_stream_t*>self._handle).write_queue_size == 0):
-
+        if self._get_write_buffer_size() == 0:
             bytes_written_ = self._try_write(buf)
 
             if bytes_written_ is None:
@@ -726,7 +726,7 @@ cdef class UVStream(UVBaseTransport):
             # buffer remaining data and send it later
 
         self._buffer_write(buf)
-        self._initiate_write()
+        self._initiate_write(True)  # skip fast path in _initiate_write
 
     def writelines(self, bufs):
         self._ensure_alive()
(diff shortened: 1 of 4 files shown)


tarasko:
This also increases aiohttp websocket echo performance by 20-25%

tarasko:
Hi @fantix, I saw you recently commited something to uvloop. Do you know if someone could take a look at this PR and maybe give a feedback?

fantix:
Yeah I'll go through all issues/PRs again and include some in the final 0.21 release, it's just all taking some time unfortunately, thanks for the PR and your kind patience!

tarasko:
Actually, this change is not very consistent with ssl write, and I was able to implement even a faster workaround for my project that doesn't touch uvloop

tarasko:
> Yeah I'll go through all issues/PRs again and include some in the final 0.21 release, it's just all taking some time unfortunately, thanks for the PR and your kind patience!

Hi @fantix, sorry to bother, I know it is open source and you're very busy. But would you have some time to look at this (and other my PRs) any time soon?

Doch88:
Can someone from the contributors have a look at this PR? It would be nice to improve the performances!

nevmerzhitsky:
@MagicStack guys please finish this CR.

pedroconceicao:
@fantix are you planning to review and release this soon? 🙏

fantix:
Sorry about the delay! I'm back on this one now.

tarasko:
> Actually instead of replicating the fast-path code, maybe we can just relax the fast-path condition, like:
> 
> ```diff
> @@ -424,7 +424,7 @@ cdef class UVStream(UVBaseTransport):
>      cdef inline _initiate_write(self):
>          if (not self._protocol_paused and
>                  (<uv.uv_stream_t*>self._handle).write_queue_size == 0 and
> -                self._buffer_size > self._high_water):
> +                (self._buffer_size > self._high_water or len(self._buffer) == 1)):
>              # Fast-path.  If:
>              #   - the protocol isn't yet paused,
>              #   - there is no data in libuv buffers for this stream,
> ```
> 
> (kind of a follow-up of [46dd8f3](https://github.com/MagicStack/uvloop/commit/46dd8f38c8ecf4e98fe3ebed9e54e5aa6f7e354a))
> 
> I believe this would have the same performance boost as this PR, yet being compatible with the current code base.

Done. 

1.  I have relaxed fast-path condition further and also removed self._buffer_size > self._high_water. Seems to be working fine.
2. I added primitive return types to _try_write and _exec_write. Previously there were PyObject_RichCompare calls for values returned by _try_write.
3. Simplied return value meaning for _try_write. Now it just returns number of bytes written or -1 in case of fatal error.

uvloop/handles/stream.pyx
```diff
@@ -681,8 +683,47 @@ cdef class UVStream(UVBaseTransport):
         if self._conn_lost:
             self._conn_lost += 1
             return
+
+        cdef ssize_t bytes_written
+
+        if self._get_write_buffer_size() == 0:
+            bytes_written_ = self._try_write(buf)
```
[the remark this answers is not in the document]
tarasko: I have removed this code completely and use _exec_write

uvloop/handles/stream.pyx
```diff
@@ -368,6 +370,8 @@ cdef class UVStream(UVBaseTransport):
             blen = py_buf.len
 
         if blen == 0:
+            if used_buf:
+                PyBuffer_Release(&py_buf)
```
fantix: Good catch!

tarasko: I added PyBuffer_Release for consistency, but I think that the outer check if blen == 0: is not really necessary.

We already filter out empty data objects twice. First in write(), second in _buffer_write(). 
We can probably remove the check in write() and leave check only in _buffer_write(). Both write and writelines rely on it.

I can do it, but I'd rather put it into a separate PR

fantix: Yeah let's merge this first.

PR #610, opened 2024-07-05
Test with Python 3.13

edgarrmondragon:
Related:

- Might be blocked by https://github.com/MagicStack/uvloop/pull/587

PR #623, opened 2024-08-28
Add cleanup_socket param on create_unix_server()

fantix:
This is derived from python/cpython#111483 but available on all Python versions with uvloop, only that it's only enabled by default for Python 3.13 and above to be consistent with CPython behavior.

[Sample CI run on 3.13](https://github.com/MagicStack/uvloop/actions/runs/10602346530/job/29384175737?pr=610)


fantix:
Merged in #610, @1st1 feel free to drop post-merge review comments.

PR #587, opened 2023-12-28
Updates for Cython3

alan-brooks:
Remove SSL depreciation warnings
buillt on debian 12.6 python-3.11 Cython-3.0.7 libuv-1.46.0

PR #634, opened 2024-10-15
Use Cython `enum` for `__PREALLOCED_BUFS`

jakirkham:
Follow up to this discussion: https://github.com/MagicStack/uvloop/pull/587#discussion_r1800388359

Simplify the code to use just a Cython `enum`. Also drop the `bint` type, which doesn't make sense given the value assigned.


jakirkham:
Thanks Fantix! 🙏

PR #638, opened 2024-10-20
test on 3.14

graingert:


Vizonex (commented):
Like the fact that my approach from winloop is being utilized but would it be smart if the install() function threw an error if 3.16+ is used?

uvloop/includes/stdlib.pxi
```diff
@@ -44,7 +44,7 @@ cdef aio_isfuture = getattr(asyncio, 'isfuture', None)
 cdef aio_get_running_loop = getattr(asyncio, '_get_running_loop', None)
 cdef aio_set_running_loop = getattr(asyncio, '_set_running_loop', None)
 cdef aio_debug_wrapper = getattr(asyncio.coroutines, 'debug_wrapper', None)
-cdef aio_AbstractChildWatcher = asyncio.AbstractChildWatcher
+cdef aio_AbstractChildWatcher = getattr(asyncio, "AbstractChildWatcher", ())
```
1st1, line 47: Why `()` and not just `None`?

1st1 (approved):
Good stuff, i've only two nit comments

PR #639, opened 2024-10-23
Fix SSLProtocol.connection_lost not being called when underlying socket is closed

cjavad:
As as a follow up to https://github.com/python/cpython/issues/118950, the fix to the issue in CPython's `asyncio` module seemed to translate quite well onto uvloop, and my failing tests effectively work both using the underlying socket streams and using higher level libraries such as aiohttp where the issue originally appeared.

As a quick description the issue occurs when an underlying socket is broken (typically accompanied by a BrokenPipe error), in uvloop this translates to the `system.write` call returning some error code which is handled in the exact same way by calling `fatal_error` as the sister implementation in asyncio. This implementation has a similar pitfall when the writer loop actively is ran in some kind of loop where the `call_soon` `connection_lost` function that is supposed to mark the transport as closed (leading to an error being thrown) never is called, leading to some pretty annoying invalid state. But these details are mostly tested using asyncio but the same symptoms and underlying reasons seems to effect this implemention after some light testing, but there might be more to the story here.

These changes are effectively a port of https://github.com/python/cpython/pull/118960 with absolutely no difference except the more thorough tests reflecting the existing tests provided.

I have initially just skipped all asyncio tests as the upstream PR has yet to be merged but most likely we will see it in a future 3.12 release and newer.

Feel free to address any issues or things i've left out from here.

### DIFF
--- a/tests/test_aiohttp.py
+++ b/tests/test_aiohttp.py
@@ -7,6 +7,7 @@ else:
     skip_tests = False
 
 import asyncio
+import os
 import sys
 import unittest
 import weakref
@@ -14,7 +15,7 @@ import weakref
 from uvloop import _testbase as tb
 
 
-class _TestAioHTTP:
+class _TestAioHTTP(tb.SSLTestCase):
 
     def test_aiohttp_basic_1(self):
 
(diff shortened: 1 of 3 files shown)


cjavad:
(It would appear the flake8 version CI/CD is using is old enough to have issues haha, locally everything is good and the error is nonsensical)

tests/test_aiohttp.py
```diff
@@ -117,8 +117,10 @@ async def stop():
         self.loop.run_until_complete(stop())
 
     def test_aiohttp_connection_lost_when_busy(self):
-        if self.implementation == 'asyncio':
-            raise unittest.SkipTest('bug in asyncio #118950 tests in CPython.')
+        if self.implementation == 'asyncio' and sys.version_info < (3, 13):
```
cjavad: This test might fail under Python <3.12.8, ==3.13.0 as the fix landed in 3.13.1 and 3.12.8, depending on how the tests are ran. 

**Edit**: Should be a non-issue though.

fantix: Ohh thanks for the info! Lemme update that too.

cjavad: Np at all, appreciate this being looked at! Let me know if there was anything else.

cjavad:
Hi again @fantix wanted to leave a question here with you, but no need to hurry getting back to me!

I was wondering what the current plan for the next release of uvloop was looking like, can we expect this fix, and others like it to land sometime soon, or would you perhaps be open to getting some more hands on deck to get to that point?

Feel more than welcome to respond here or reach out to me via other channels.

Thanks again, and enjoy the time between now and whenever you get back to me haha :)

Issue #506, opened 2022-09-28
unable to perform operation on <WriteUnixTransport closed=True reading=False 0x7f16825f8ba0


m0hammadhossein:
* **uvloop version**: 0.17.0
* **Python version**: 3.10.7
* **Platform**: ubuntu 20.04
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: yes
* **Does uvloop behave differently from vanilla asyncio? How?**: yes

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/pytgcalls/methods/groups/join_group_call.py", line 275, in internal_sender
    await self._binding.send(request)
  File "/usr/local/lib/python3.10/dist-packages/pytgcalls/binding.py", line 212, in send
    await self._send({
  File "/usr/local/lib/python3.10/dist-packages/pytgcalls/binding.py", line 221, in _send
    self._js_process.stdin.write(
  File "/usr/lib/python3.10/asyncio/streams.py", line 326, in write
    self._transport.write(data)
  File "uvloop/handles/stream.pyx", line 674, in uvloop.loop.UVStream.write
  File "uvloop/handles/handle.pyx", line 159, in uvloop.loop.UVHandle._ensure_alive
RuntimeError: unable to perform operation on <WriteUnixTransport closed=True reading=False 0x7f16825f8ba0>; the handler is closed

```

kekekekule:
got the same issue with 0.16

andrmueller:
got the issue when using aiomysql, probably the [issue](https://github.com/aio-libs/aiomysql/issues/966) people are facing with aiomysql and uvloop seems to be related to this issue here.

Maybe the information posted in the aiomysql issue helps if someone takes this over.

Kludex:
@1st1 We talked at PyCon US this year. 👋 I said I would create an issue about it... The issue is this one.

Let me share an MRE:

```py
import asyncio


class EchoServerProtocol(asyncio.Protocol):
    def connection_made(self, transport: asyncio.Transport):  # type: ignore
        self.transport = transport

    def data_received(self, data: bytes):
        message = data.decode()
        print("Data received: {!r}".format(message))

        self.transport.close()

        # We introduce a short wait to ensure closure processing
        asyncio.ensure_future(self.write_after_closure(data))

    async def write_after_closure(self, data: bytes):
        print(self.transport)
        # uvloop: <TCPTransport closed=True reading=False 0x153e33aa0>
[...]

Issue #645, opened 2024-11-20
Information exposure bug: response leaking to wrong requests in `uvicorn`-based server under heavy load


MarkusSintonen:
* **uvloop version**: 0.21.0
* **Python version**: 3.12.6
* **Platform**: `uname_result(system='Linux', node='xxx', release='6.1.109', version='#1 SMP PREEMPT_DYNAMIC Thu Sep 19 22:28:35 UTC 2024', machine='x86_64')`
* **Can you reproduce the bug with `PYTHONASYNCIODEBUG` in env?**: No
* **Does uvloop behave differently from vanilla asyncio? How?**: Yes, see below

We encountered a very nasty data leakage bug with `uvloop`. It seems to be leaking responses to incorrect requests in a `uvicorn` based server under a heavy load. This caused some users data to leak into requests of other users leading to a incorrect information exposure.

This issue seems to be caused solely by `uvloop` as removal of it fixed the issue. Relying on vanilla asyncio does not have the same issue.

The issue happened under high load situations. The service processes some 800 million requests per day but 100 requests in a day had the wrong responses from other concurrently happening requests. It seemed to happen in a situations when there is a higher load. Also the issue sometimes correlated with other issues we experienced with `uvloop`:
- https://github.com/aio-libs/aiohappyeyeballs/issues/93
- https://github.com/MagicStack/uvloop/issues/506

Sometimes we observed these strange and bad looking `RuntimeError`s coming from depths of `uvloop` at about same time as we saw the incorrect responses coming from requests. But this did not happen always in correlation. (The above errors also got fixed by removal of `uvloop`).

I haven't been able to reproduce this as it seems to only happen under heavy load situations and rarely enough (but still bad to leak information).

It seems `uvloop` might have some major issues in its TCP/socket/stream implementation that it tries to some times use incorrect already used socket like the RuntimeErrors would hint. Not sure is it actually related to the data leakage issue.

MarkusSintonen:
> Is this duplicated of #631 ?

I dont know, that seems to be about `handling multiple requests for the same path the response took longer than expected to arrive at the client.`. So something about performance. But it could be related.

MarkusSintonen:
[...]

PR #646, opened 2024-11-25
Fix for unexpected socket closures and data leakage under heavy load

todddialpad:
This is to address issue #645 and in aiohttp/aiohappyeyeballs#93 and aiohttp/aiohappyeyeballs#112

### DIFF
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -1877,6 +1877,7 @@ cdef class Loop:
             AddrInfo ai_local = None
             AddrInfo ai_remote
             TCPTransport tr
+            int sockfd
 
             system.addrinfo *rai = NULL
             system.addrinfo *lai = NULL
@@ -2060,8 +2061,10 @@ cdef class Loop:
             waiter = self._new_future()
             tr = TCPTransport.new(self, protocol, None, waiter, context)
             try:
+                # Take ownership of the file descriptor
+                sockfd = sock.detach()
                 # libuv will make socket non-blocking
-                tr._open(sock.fileno())
+                tr._open(sockfd)
                 tr._init_protocol()
                 await waiter
             except (KeyboardInterrupt, SystemExit):
@@ -2074,8 +2077,6 @@ cdef class Loop:
                 # `_close()` before it is fine.
                 tr._close()
                 raise
-
-            tr._attach_fileobj(sock)
 
         if ssl:
             app_transport = protocol._get_app_transport(context)


uvloop/loop.pyx
```diff
@@ -2060,8 +2061,10 @@ cdef class Loop:
             waiter = self._new_future()
             tr = TCPTransport.new(self, protocol, None, waiter, context)
             try:
+                # Take ownership of the file descriptor
+                sockfd = sock.detach()
                 # libuv will make socket non-blocking
-                tr._open(sock.fileno())
+                tr._open(sockfd)
```
1st1, line 2067: The approach looks correct -- but I'm wondering how vanilla asyncio handles the same thing?

todddialpad:
> @todddialpad very nice! Do you know does it help with the other issue #506 which seems to be also related to incorrect sharing of sockets etc?
> 
> Any possibility to add some test here?

I am trying to get a stable test. It is tricky because it is a race condition, if my guess is correct. I think it is a race if TLS negotiation during a call to `loop.create_connection` with an explicit socket is cancelled, and a subsequent incoming connection is accepted before the `CancelledError` is propagated. I think both `libuv` (or `uvloop`) first and `aiohttp` second close the underlying file descriptor.

So if this is the case, I don't think this will fix issue #506 , which could be a similar but different root cause.

MarkusSintonen:
> So if this is the case, I don't think this will fix issue https://github.com/MagicStack/uvloop/issues/506 , which could be a similar but different root cause.

Ok I see, the linked issue was also concerning as it looked as it was trying to write data into some incorrect socket. The error was also something we observed at similar time instances when we observed the response data getting leaked to incorrect requests. But we dont know is that issue actually related to the data leakage or just something else. (These RuntimeErrors dont happen with vanilla asyncio)

uvloop/loop.pyx
todddialpad: I think vanilla asyncio has an easier problem in that it can just have python sockets "all the way down", so just let reference counting take care of cleanup, while here we need to manage the disconnect with `libuv` dealing in file descriptors. I am suspecting there is some error handling path where a file descriptor is closed while the python socket object remains alive and not detached, so when it **is** finally closed, it messes up any new socket that happens to have the same file descriptor.
e.g. create socket `s`, call a loop method passing in an explicit socket, <bad error path which will end with `sock.close()`> overlapping with an `.accept`. I think the `.accept` never results in a python socket object being created.

So with the methods accepting `sockets` and other methods that internally work directly in file descriptors can there be a discrepancy?

todddialpad:
> > @todddialpad very nice! Do you know does it help with the other issue #506 which seems to be also related to incorrect sharing of sockets etc?
> > Any possibility to add some test here?
> 
> I am trying to get a stable test. It is tricky because it is a race condition, if my guess is correct. I think it is a race if TLS negotiation during a call to `loop.create_connection` with an explicit socket is cancelled, and a subsequent incoming connection is accepted before the `CancelledError` is propagated. I think both `libuv` (or `uvloop`) first and `aiohttp` second close the underlying file descriptor.
> 
> So if this is the case, I don't think this will fix issue #506 , which could be a similar but different root cause.

I still haven't been able to isolate a standalone, self-contained test. The test environment in which I generated the same error we see in production involves 2 VMs with significant network latency between them. The first of the VMs is just a web server, the second is a web server that accepts requests, and then makes outgoing client requests (using aiohttp) to the first webserver with TLS and a short timeout (around 1 second).

With this setup, I quite reliably get a failure within 250 connections. When I run with this patch applied, I have never had a failure in 20,000 connections.

We have also run this in our production environment. When we first encountered this failure, we hit it within 1 hour of using aiohttp >= 3.10. Since running with this patch we have been running for 5 days with no failures.

todddialpad:
Is accepting this blocked on the tests that are failing? I don't think those failures are related to this change, as they are also failing for PR #644, which is solely a documentation change.

I looked at the test logs and I would guess that a dependency is causing the changed results. Related to this, I notice that in the failing tests, and alpha release of Cython 3.1 is being used (`Using cached Cython-3.1.0a1-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata`). Is this intentional?
In the last test run that passed, the release version was used (`Using cached Cython-3.0.11-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata`)

AntonArsentiev:
Hello everyone. Did i think right that this MR fix issues below?

"RuntimeError: File descriptor 2877 is used by transport <TCPTransport closed=False reading=True 0x55b8dc9baa90>"

AntonArsentiev:
Hello everyone :) 

Like many other users of this library, I would be happy for this fix to be implemented in one of the upcoming releases. 
Could you figure out roughly when to expect this fix?

Dreamsorcerer:
@1st1 @fantix Is anyone available to merge and release this? We're getting asked to put workarounds into aiohttp to deal with this, would be nice to have this fixed here instead.

bdraco:
We added a workaround for this issue in https://github.com/aio-libs/aiohttp/pull/10464 but its causing using with asyncio SelectorEventLoop https://github.com/aio-libs/aiohttp/issues/10617 so we will likely be reverting it and waiting for this PR instead

webknjaz:
Hey @elprans @1st1, do you think you'll be able to spare a minute to get this in?

1st1:
> Hey @elprans @1st1, do you think you'll be able to spare a minute to get this in?

Sorry, @fantix and I will be going through this PR and others this week.

jperezr21:
Hey! I noticed that `aiohttp 3.11.14` has been yanked. For those of us using `uvloop` and `aiohttp` and running into the `File descriptor 91 is used by transport` error, do you happen to know if there’s a temporary workaround or a specific combination of versions we can pin to in the meantime? Totally understand if we need to wait for this to be merged, just trying to keep things running smoothly in the short term. Thanks a lot!

Dreamsorcerer:
You can pin to a yanked version.

jugalshah291:
I folks 
We are also running into below issue, 
```
File descriptor 91 is used by transport
```

Wonder if their is a fix

Our setup
```
aiohappyeyeballs==2.6.1
aiohttp==3.11.18
aiohttp-cors==0.8.1
uvloop==0.21.0
```

jperezr21:
Pinning to `aiohttp==3.11.14` solved it for me

bdraco:
> Pinning to `aiohttp==3.11.14` solved it for me

Just a heads-up: if you're using the default asyncio event loop (typically `SelectorEventLoop`), pinning to `aiohttp==3.11.14` may introduce other issues due to some side effects in that version. If you're using `uvloop` exclusively, it's likely fine.

Ideally, we were hoping this PR would be merged to avoid relying on workarounds in `aiohttp` as we've already been down that road, had to revert, and don’t want a repeat. Unfortunately, this PR seems to have stalled.

jugalshah291:
Can we please prioritize this PR?

todddialpad:
@fantix thanks for having a look at this. I am not sure what to make of the test failures. The failures seem to be all related to Unix transports and subprocess transports. The PR only should affect TCP transports. I'm trying to repro. My dev environment is Ubuntu / py3.12, and the tests that are failing here are passing there. For example:
```
test_process_send_signal_1 (test_process.Test_UV_Process.test_process_send_signal_1) ... ok
test_process_streams_basic_1 (test_process.Test_UV_Process.test_process_streams_basic_1) ... ok
test_process_streams_devnull (test_process.Test_UV_Process.test_process_streams_devnull) ... ok
test_process_streams_pass_fds (test_process.Test_UV_Process.test_process_streams_pass_fds) ... ok
```

Do you have any ideas on how to proceed?

fantix:
They are breaking in the debug build, maybe try this:

https://github.com/MagicStack/uvloop/blob/96b7ed31afaf02800d779a395591da6a2c8c50e1/.github/workflows/tests.yml#L68-L71

todddialpad:
> * Tests / test (3.12, macos-latest) (pull_request)

Yes, I get the failures with the debug build, good eye. Thanks.

I have instrumented the changed code, and in a failing test, the modifications never even run (which makes sense since the test isn't creating any TCP connections).

I have built without this patch, and still see the failures with the debug build.
```
git clone --recursive https://github.com/magicstack/uvloop.git uvloop.official
cd uvloop.official/
python3 -m venv uvloop-dev
source uvloop-dev/bin/activate
pip install -e .[dev]
pip install psutil
make debug
make test
```

```
======================================================================
FAIL: test_process_streams_pass_fds (test_process.Test_UV_Process.test_process_streams_pass_fds) [Alive handle after test] (handle_name='UVProcessTransport')
----------------------------------------------------------------------
  File "uvloop.official/uvloop/_testbase.py", line 142, in tearDown
    self.assertEqual(
AssertionError: 1 != 0 : alive UVProcessTransport after test

```

So, could an upstream dependency have broken the debug build?

todddialpad:
> So, could an upstream dependency have broken the debug build?

Since the last successful test run, the following upstream dependencies have changed:
```
Cython-3.1.0 (was 3.0.12)
aiohttp-3.11.18 (was 3.11.16)
frozenlist-1.6.0 (was 1.5.0)
mypy_extensions-1.1.0 (was 1.0.0)
setuptools-80.3.1 (was 78.1.0)
```

I rebuilt using Cython-3.0.12 and the tests passed.

Would a manual execution of the tests on the main branch still pass (assuming it will grab Cython 3.1.0)?

todddialpad:
> Would a manual execution of the tests on the main branch still pass (assuming it will grab Cython 3.1.0)?

I forked the main branch and tried running the tests. It fails with Cython 3.1.0. I pinned Cython to < 3.1.0 and the tests pass. I included this PR, and with the pinned Cython, all tests pass.

So I believe this PR could be merged. I created an issue for Cython 3.1.0 #677 .

jugalshah291:
Hi checking back on this, any ETA on when it would be merged

GabrielSalla:
Following this PR waiting for the fix

jugalshah291:
Hi folks any ETA on this

jugalshah291:
Hi Folks
can someone provide an update on this
We are facing an issue where we suddenly start seeing `File descriptor 91 is used by transport` on a running process and while the error happens the process is not able to serve traffic. This is impacting our service stability

yybdyybd:
> Hi FolksHi Folks 大家好 can someone provide an update on this can someone provide an update on this有人能提供一下这个的最新情况吗 We are facing an issue where we suddenly start seeing We are facing an issue where we suddenly start seeing 我们遇到了一个问题，即突然开始看到`File descriptor 91 is used by transport` on a running process and while the error happens the process is not able to serve traffic. This is impacting our service stability on a running process and while the error happens the process is not able to serve traffic. This is impacting our service stability在一个正在运行的进程上，当错误发生时，该进程无法处理流量。这正影响着我们的服务稳定性。

Hello, I also have this problem. This problem occurs when calling a third-party interface times out and is in a high-concurrency scenario. Have you solved it? Please advise.

Dreamsorcerer:
> Have you solved it? Please advise.

Uninstall uvloop? Several users have reported that the performance difference is small today, so if it's breaking your application...

x0day:
gently ping @fantix @1st1

fantix:
This is now fixed in #740 with a different approach, which preserved the whole `transport._attach_fileobj()` setup for asyncio compatibility (see also python/asyncio#449). #740 also fixed the same issue on UDS, and added simulation tests.

PR #661, opened 2025-03-22
fix eof_received

dennissheng:
I think it's not right to shutdown when eof received.  It's better to wait for app resume when app reading paused. I wrote a test case(test_close_notify.py) for this situation. Even when there are no app reading pause, reading from buffer is needed before shutdown, cause sometimes ssl eof is in the same package of tcp fin. So when connection is closing, reading buffer is still needed to parse the EOF alert message in this package.
image

### DIFF
--- a/uvloop/sslproto.pyx
+++ b/uvloop/sslproto.pyx
@@ -402,12 +402,17 @@ cdef class SSLProtocol:
             if self._state == DO_HANDSHAKE:
                 self._on_handshake_complete(ConnectionResetError)
 
-            elif self._state == WRAPPED or self._state == FLUSHING:
-                # We treat a low-level EOF as a critical situation similar to a
-                # broken connection - just send whatever is in the buffer and
-                # close. No application level eof_received() is called -
-                # because we don't want the user to think that this is a
-                # graceful shutdown triggered by SSL "close_notify".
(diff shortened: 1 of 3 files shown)


1st1:
@fantix this is your area of expertise :)

uvloop/sslproto.pyx
```diff
@@ -597,6 +599,11 @@ cdef class SSLProtocol:
         if close_notify:
             self._call_eof_received(context)
 
+    cdef _do_read_flush(self):
+        self._do_read()
+        self._set_state(SHUTDOWN)
+        self._on_shutdown_complete(None)
```
entelligence-ai-reviews, line 605: **Correctness**: The `_do_read_flush` method is called in the `FLUSHING` state but doesn't handle the case when app reading is paused, which could lead to data loss when the client pauses reading.


<details>
<summary><strong>📝 Committable Code Suggestion</strong></summary>

> ‼️ Ensure you review the code suggestion before committing it to the branch. Make sure it replaces the highlighted code, contains no missing lines, and has no issues with indentation.

```suggestion
    cdef _do_read_flush(self):
        if not self._app_reading_paused:
            self._do_read()
            self._set_state(SHUTDOWN)
            self._on_shutdown_complete(None)
```
</details>

PR #679, opened 2025-05-12
Fix cpython imports

Vizonex:
C imports from cpython alone are deprecated and warned against & discouraged. I have already migrated and made these changes to [the Winloop Project](https://github.com/Vizonex/Winloop/commit/c422d4fedd18656691469950a01198d1b346e41b) Know that our code-spaces are a bit different from this one due to different operating systems being used. However I did my best to keep the imports in alphabetical order.


Vizonex:
I Guess another thing I accidently included was the redirection to the winloop project if someone using windows were to attempt to install this project.

PR #693, opened 2025-10-11
add free-threading support

kumaraditya303:
This adds free-threading support by adding `freethreading_compatible=True` in `loop.pyx` and also enables CI and wheels for 3.14t.

cc @fantix

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ docs = [
 requires = [
     "setuptools>=60",
     "wheel",
-    "Cython~=3.0",
+    "Cython~=3.1",
 ]
 build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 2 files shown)


Vizonex:
Good News: I have successfully mirrored the changes from here, winloop 0.3.0 will release soon since I now obey how version numbers actually work and since the `__init__.py` script for winloop significantly changed. Thanks for your help again. 

Bad News: uvloop also seems there's some rounding bugs with macos from the tests that I've seen I am also question why my library has the same problem. But maybe if we find a way to fix it I'll try doing the same with winloop's end. Rounding has always been a bit of a problem for a while (not sure why exactly).

fantix:
Yeah, it's definitely getting more flaky to run tests, and there're a few warnings to take care of. Let's release for Python 3.14 first, and get back to those as follow-ups.

Vizonex:
> Yeah, it's definitely getting more flaky to run tests, and there're a few warnings to take care of. Let's release for Python 3.14 first, and get back to those as follow-ups.

Good idea.

PR #698, opened 2025-10-13
Start applying ruff rules

DimitriPapadopoulos:
I have selected ruff rules that seem to make most sense. My idea is to suggest switching from flake8 to ruff.


uvloop/loop.pyi
```diff
@@ -1,41 +1,35 @@
...
-    Union,
     overload,
 )
 
 _T = TypeVar('_T')
-_Context = Dict[str, Any]
+_Context = dict[str, Any]
 _ExceptionHandler = Callable[[asyncio.AbstractEventLoop, _Context], Any]
-_SSLContext = Union[bool, None, ssl.SSLContext]
+_SSLContext = bool | ssl.SSLContext | None
```
fantix, line 18: Okay this would require we drop support for 3.8 and 3.9, and we should. Lemme do that with another PR.

fantix (commented):
Thank you for the PR! Yes, lets do it.

Do you still have your `[tool.ruff]` section in `pyproject.toml`? We should include that too I think.

tests/test_process_spawning.py
```diff
@@ -100,8 +100,8 @@ def spawn_process():
                 t.start()
                 t.join(timeout=10.0)
                 if t.is_alive():
-                    raise Exception('process freeze detected at {}'
-                                    .format(iteration))
+                    raise Exception(f'process freeze detected at {iteration}'
+                                    )
```
fantix:
```suggestion
                    raise Exception(f'process freeze detected at {iteration}')
```

uvloop/loop.pyi
DimitriPapadopoulos: Strange, Ruff is supposed to read this:
https://github.com/MagicStack/uvloop/blob/173e88cfcdae8edc1763dd9915087f5136b568aa/pyproject.toml#L5
and adapt to whatever version of Python defined by `requires-python`. Not sure what happened here 🤔

* [Inferring the Python version](https://docs.astral.sh/ruff/configuration/#inferring-the-python-version)

tests/test_process_spawning.py
DimitriPapadopoulos: A formatter such as `ruff format` would fix that, but let's leave that for another PR. i'll fix it manually here.

DimitriPapadopoulos:
I used `--extend-select`/`--select` to apply a single rule per commit, in case you wouldn't like a specififc rule. But I can add a `[tool.ruff]` section, no problem.

fantix:
> I used `--extend-select`/`--select` to apply a single rule per commit, in case you wouldn't like a specififc rule. But I can add a `[tool.ruff]` section, no problem.

Ohh rule per commit is definitely great! It's just so that future code could simply apply the same rules, adding a `[tool.ruff]` would be necessary? But we can do that in a second PR to replace `test_flake8` with `test_ruff`

uvloop/loop.pyi
fantix: Yeah anyways, the fix is good! :heart:

fantix: (It's probably because it's a `pyi` file instead of `py`; `X|Y` grammar in `pyi` files is actually supported by mypy/pyright on Python 3.9.)

fantix (approved):

PR #704, opened 2025-10-22
call_later should always return TimerHandle

jensbjorgensen:
Python's builtin asyncio Loop always returns a TimerHandle from call_later/call_at, however current uvloop will instead fall back to call_soon if delay==0. It seems desirable that uvloop be api-compatible. I'm sure there's a little performance improvement from call_soon if you can use it, but end-users call always use call_soon anyway if they know the delay is zero.

Demonstration:

```
import asyncio, uvloop
asyncio.new_event_loop().call_later(0, lambda : None).when()
uvloop.new_event_loop().call_later(0, lambda : None).when()

```


jensbjorgensen:
oops sorry about the mis-formatting in the first commit, that's fixed now. I see macos tests failing, but this doesn't seem to be due to me.

PR #721, opened 2026-01-19
Fix race conditions on `loop._ready_len`

### LINKED ISSUE #720: `loop.call_soon_threadsafe()` behaves as non-thread-safe in free-threading
When I tried running [rogerbinns/python-async-bench](https://github.com/rogerbinns/python-async-bench) on free-threaded Python 3.14.2 (Linux, dual-core), I found it getting stuck on the uvloop (0.22.1) tests. Below is the code to reproduce the issue and its possible output.

```python
#!/usr/bin/env python3

import asyncio
import threading
import time

from queue import SimpleQueue

import uvloop

START_TIME = time.monotonic()


def callback(future):
    future.set_result(True)
    print(f"time: {time.monotonic() - START_TIME:.10f} (callback)")


def notify_all(loop, queue):
    while (future := queue.get()) is not None:
        loop.call_soon_threadsafe(callback, future)


async def main():
    queue = SimpleQueue()
    thread = threading.Thread(target=notify_all, args=[loop, queue])
    thread.start()

    try:
        async with asyncio.timeout(60):
            while True:
                future = loop.create_future()

                print(f"time: {time.monotonic() - START_TIME:.10f}")

                try:
                    async with asyncio.timeout(3):
                        outer_future = asyncio.shield(future)
                        queue.put(future)
                        await outer_future
                except TimeoutError:
                    print("[timeout] future:", future)
                    raise
    except TimeoutError:
        pass
    finally:
        queue.put(None)

    await asyncio.to_thread(thread.join)


if __name__ == "__main__":
    uvloop.run(main())
```

```
...
time: 0.1763970610
time: 0.1764364019 (callback)
time: 0.1764616738
time: 0.1765011698 (callback)
time: 0.1765267970
time: 3.1786725279 (callback)
[timeout] future: <Future finished result=True>
```

This looks like a race condition. As we can see, the handle was successfully added, but was only processed by the `_on_idle()` method after the timeout. It is probably related to the fact that access to the `_ready_len` attribute is performed non-atomically from different threads (or, in a simpler case, [`self._ready_len = len(self._ready)`](https://github.com/MagicStack/uvloop/blob/836e3b2eb7c0dbe591848c32fe9c4b0f729de760/uvloop/loop.pyx#L484-L486) from the main thread competes with [`self._ready_len += 1`](https://github.com/MagicStack/uvloop/blob/836e3b2eb7c0dbe591848c32fe9c4b0f729de760/uvloop/loop.pyx#L675) from the worker thread). This also applies to [`loop.run_in_executor()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor) and [`asyncio.to_thread()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread) running on top of it, but it is much more difficult to reproduce the issue with them (due to the greater delay between threads).

Related: #408.

x42005e1f:
This is a simple PR that fixes #720 by removing `loop._ready_len` in favor of `len(loop._ready)`. The attribute was probably added as a premature optimization (it does not affect performance even in edge cases), but it leads to race conditions both between worker threads calling `loop.call_soon_threadsafe()` and between any such worker thread and the event loop, since `loop._ready_len` is updated in a non-thread-safe manner (free-threading).

### DIFF
--- a/uvloop/loop.pxd
+++ b/uvloop/loop.pxd
@@ -50,7 +50,6 @@ cdef class Loop:
         object _default_executor
         object _ready
         set _queued_streams, _executing_streams
-        Py_ssize_t _ready_len
 
         set _servers
 
--- a/uvloop/loop.pyx
+++ b/uvloop/loop.pyx
@@ -181,7 +181,6 @@ cdef class Loop:
         self._queued_streams = set()
         self._executing_streams = set()
         self._ready = col_deque()
-        self._ready_len = 0
 
         self.handler_async = UVAsync.new(
             self, <method_t>self._on_wake, self)
@@ -440,7 +439,7 @@ cdef class Loop:
             self.handler_async.send()
 
     cdef _on_wake(self):
-        if ((self._ready_len > 0 or self._stopping) and
+        if ((len(self._ready) > 0 or self._stopping) and
                 not self.handler_idle.running):
             self.handler_idle.start()
 
@@ -481,8 +480,7 @@ cdef class Loop:
         if len(self._queued_streams):
             self._exec_queued_writes()
 
-        self._ready_len = len(self._ready)
-        if self._ready_len == 0 and self.handler_idle.running:
+        if len(self._ready) == 0 and self.handler_idle.running:
             self.handler_idle.stop()
 
         if self._stopping:
@@ -570,7 +568,6 @@ cdef class Loop:
         for cb_handle in self._ready:
             cb_handle.cancel()
         self._ready.clear()
-        self._ready_len = 0
 
         if self._polls:
             for poll_handle in self._polls.values():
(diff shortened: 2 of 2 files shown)


x42005e1f:
The test failure looks like a load on CI/CD. Compared to #719, some jobs run faster, some run at the same speed, and some run slower. The problem with `test_process_streams_redirect()` is that it also captures asyncio messages in stderr that a task's callback is taking too long (probably due to slow context switching between threads/processes under high load on the system scheduler?), so it is flaky. The warning is [logged from `_on_idle()`](https://github.com/MagicStack/uvloop/blob/master/uvloop/loop.pyx#L458-L469), so it is unrelated to the changes in this PR.

Issue #702, opened 2025-10-17
`asyncio.get_event_loop` raises error instead of spawning new loop when no loop is running.


anonymousx97:
```python
>>> import uvloop, asyncio
>>> asyncio.get_event_loop()
<_UnixSelectorEventLoop running=False closed=False debug=False>
>>> uvloop.install()
>>> asyncio.get_event_loop()
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    asyncio.get_event_loop()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/uvloop/__init__.py", line 206, in get_event_loop
    raise RuntimeError(
    ...<2 lines>...
    )
RuntimeError: There is no current event loop in thread 'MainThread'.
>>>
```

As you can see in the snippet above when run on un-patched asyncio it's returning a new loop. But uvloop raises error because of the [method override](https://github.com/MagicStack/uvloop/blob/74f4c96d3fc5281b1820491d2568de771ea7851b/uvloop/__init__.py#L200) which was not there in previous version inheriting the original new loop spawning function.

fantix:
Roger; I'm looking into this

andylin-hao:
Same issue here. It looks like 0.22.x's `EventLoopPolicy` overrides its base class `asyncio.events.BaseDefaultEventLoopPolicy`'s `get_event_loop` function, and removes the creating new loop behavior.

DiegoD94:
Same here, multiple project, including sglang and verl is impacted

shadchin:
This diff is fixing for me (copy from BaseDefaultEventLoopPolicy):
```diff
--- a/uvloop/__init__.py    (index)
+++ b/uvloop/__init__.py    (working tree)
@@ -193,6 +193,7 @@ def __getattr__(name: str) -> _typing.Any:

         class _Local(threading.local):
             _loop: _typing.Optional[_AbstractEventLoop] = None
+            _set_called: bool = False

         def __init__(self) -> None:
             self._local = self._Local()
@@ -202,6 +203,28 @@ def __getattr__(name: str) -> _typing.Any:

             Returns an instance of EventLoop or raises an exception.
             """
+            if (self._local._loop is None and
+                    not self._local._set_called and
+                    threading.current_thread() is threading.main_thread()):
+                stacklevel = 2
+                try:
+                    f = _sys._getframe(1)
+                except AttributeError:
+                    pass
[...]

PR #723, opened 2026-01-21
Create loop in `get_event_loop` if needed

zvyn:
This fixes issue #702.

### DIFF
--- a/uvloop/__init__.py
+++ b/uvloop/__init__.py
@@ -1,4 +1,5 @@
 import asyncio as __asyncio
+from types import FrameType
 import typing as _typing
 import sys as _sys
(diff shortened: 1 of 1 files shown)


ypnos:
This behavior is deprecated since Python 3.13: https://docs.python.org/3.13/library/asyncio-eventloop.html#asyncio.get_event_loop

zvyn:
> This behavior is deprecated since Python 3.13: https://docs.python.org/3.13/library/asyncio-eventloop.html#asyncio.get_event_loop

Thanks for pointing that out @ypnos, I changed the PR to follow the asyncio behavior more closely.

PR #727, opened 2026-02-11
Fix CI

patrickelectric:


patrickelectric:
@fantix this fixes the tests for all python versions except 3.14t, no idea why.

pyproject.toml
```diff
@@ -45,6 +45,7 @@ test = [
     'mypy>=0.800',
 ]
 dev = [
+    'packaging',
```
fantix, line 48: I think we also need this in `build-system.requires`, so that people who install from the src tarball overriding `cython-always`, would succeed. It's a marginal case though.

fantix (approved):

pyproject.toml
patrickelectric: I just updated as requested, sorry for the delay

fantix:
Thank you! This is now merged in #742 and will be mentioned in release log.

PR #734, opened 2026-03-13
add missing 'context' parameter to Loop.create_task stub

eyetheekh:
This PR adds the missing context parameter to the create_task type stub in uvloop/loop.pyi. The implementation in loop.pyx already handles the context argument, but was missing in the loop.pyi.


eyetheekh:
The CI failure is unrelated to the changes in this PR , caused by setuptools depricating pkg_resources in versions above 80.10.2

fantix (approved):

PR #740, opened 2026-03-23
Detach socket on create_connection cancellation to prevent fd double-close

### LINKED ISSUE #738: create_connection(sock=sock) does not detach socket on cancellation, causing fd double-close
## Summary

When `loop.create_connection(protocol_factory, sock=sock)` is cancelled (e.g. by a connect timeout), the passed-in Python `socket` object is never detached. Its `__del__` later closes whatever fd the OS recycled into that slot, silently corrupting an unrelated transport.

Standard asyncio is not affected — its transport directly owns the Python socket and properly closes it on the error path, setting `fileno()` to `-1`.

## Root cause

In `loop.pyx` (the `sock is not None` branch of `create_connection`):

```python
# uvloop/loop.pyx ~line 2060
waiter = self._new_future()
tr = TCPTransport.new(self, protocol, None, waiter, context)
try:
    tr._open(sock.fileno())      # libuv takes over the fd
    tr._init_protocol()          # _transports[fd] = tr
    await waiter                 # ← cancellation point
except BaseException:
    tr._close()                  # libuv closes fd, but sock doesn't know
    raise

tr._attach_fileobj(sock)         # only reached on success — calls sock.detach() later via _close()
```

On the **success path**, `_attach_fileobj` saves a reference to `sock`. When the transport is later closed, `_close()` calls `sock.detach()`, which sets the socket's internal fd to `-1` so GC won't double-close it.

On the **error/cancellation path**, `_attach_fileobj` is never called → `sock.detach()` never happens → the Python socket still believes it owns the (now-recycled) fd number → `sock.__del__()` closes an unrelated fd.

## Reproduction

```python
import asyncio
import socket
import uvloop

async def test(loop):
    srv = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    srv.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    srv.bind(("127.0.0.1", 0))
    srv.listen(1)
    addr = srv.getsockname()

    client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    client.setblocking(False)
    try:
        client.connect(addr)
    except BlockingIOError:
        pass
    await asyncio.sleep(0.01)

    task = asyncio.ensure_future(loop.create_connection(asyncio.Protocol, sock=client))
    await asyncio.sleep(0)
    task.cancel()
    try:
        await task
    except asyncio.CancelledError:
        pass

    print(f"uvloop:  sock.fileno() = {client.fileno()}")  # NOT -1 → bug
    srv.close()
    try:
        client.close()
    except OSError:
        pass

# uvloop — socket NOT detached
uvloop.install()
loop = asyncio.new_event_loop()
loop.run_until_complete(test(loop))
loop.close()

# standard asyncio — socket properly detached
asyncio.set_event_loop_policy(None)
loop = asyncio.new_event_loop()
loop.run_until_complete(test(loop))
loop.close()
```

Output:
```
uvloop:  sock.fileno() = 13    ← should be -1
asyncio: sock.fileno() = -1    ← correct
```

## Impact

Under high concurrency with `force_close=True` (aiohttp) and intermittent connection timeouts, the GC'd socket closes a live transport's fd. This manifests as:

```
RuntimeError: File descriptor 1768 is used by transport <TCPTransport closed=False reading=True 0x...>
```

followed by `Assertion failed: ok (src/mailbox.cpp:72)` → SIGABRT, crashing the process.

## Suggested fix

Detach the socket in the error path:

```python
except BaseException:
    tr._close()
    if sock is not None:
        sock.detach()
    raise
```

This matches the semantics of standard asyncio, where the transport always takes full ownership of the socket.

## Environment

- uvloop 0.21.0
- Python 3.12
- Linux 5.15

junjzhang:
## Summary

Fixes #738.

When `create_connection(sock=sock)` or `create_unix_connection(sock=sock)` is cancelled or raises, `tr._close()` closes the fd via libuv but the Python socket object still believes it owns that fd. Its `__del__` later closes whatever fd the OS recycled into that slot, silently corrupting an unrelated transport.

This adds `sock.detach()` on the error path of both `create_connection` and `create_unix_connection`, matching the semantics of standard asyncio where the transport always takes full ownership of the socket.

## Changes

- `loop.pyx`: Call `sock.detach()` after `tr._close()` in the `except BaseException` block for both TCP and Unix `sock=` paths
- `test_tcp.py`: Add `test_create_connection_sock_cancel_detaches` verifying that `sock.fileno() == -1` after cancellation

### DIFF
--- a/tests/test_tcp.py
+++ b/tests/test_tcp.py
@@ -786,7 +786,7 @@ class _TestTCP:
 
         async def test():
             srv = await asyncio.start_server(
-                lambda r, w: None,
+                lambda r, w: w.close(),
                 '127.0.0.1', 0,
                 family=socket.AF_INET)
             addr = srv.sockets[0].getsockname()
--- a/tests/test_tcp.py
+++ b/tests/test_tcp.py
@@ -1181,6 +1181,32 @@ class Test_UV_TCP(_TestTCP, tb.UVTestCase):
         # let it close
         self.loop.run_until_complete(asyncio.sleep(0.1))
 
+    def test_create_connection_sock_cancel_detaches(self):
+        async def client(addr):
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            sock.setblocking(False)
+            try:
+                sock.connect(addr)
+            except BlockingIOError:
+                pass
+            await asyncio.sleep(0.01)
+
+            task = asyncio.ensure_future(
+                self.loop.create_connection(asyncio.Protocol, sock=sock))
+            await asyncio.sleep(0)
+            task.cancel()
+            with self.assertRaises(asyncio.CancelledError):
+                await task
+
+            # After cancellation the socket must be detached (fd == -1)
+            # so that its __del__ won't close a recycled fd.
+            self.assertEqual(sock.fileno(), -1)
+
+        with self.tcp_server(lambda sock: sock.recv_all(1),
+                             max_clients=1,
+                             backlog=1) as srv:
+            self.loop.run_until_complete(client(srv.addr))
+
     @unittest.skipUnless(hasattr(socket, 'AF_UNIX'), 'no Unix sockets')
     def test_create_connection_wrong_sock(self):
         sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
--- a/tests/test_tcp.py
+++ b/tests/test_tcp.py
@@ -737,6 +737,130 @@ class _TestTCP:
         with s1, s2:
             loop.run_until_complete(test())
 
+    def test_create_connection_sock_cancel_detaches(self):
+        async def client(addr):
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            sock.setblocking(False)
+            try:
+                sock.connect(addr)
+            except BlockingIOError:
+                pass
+            await asyncio.sleep(0.01)
+
+            task = asyncio.ensure_future(
+                self.loop.create_connection(asyncio.Protocol, sock=sock))
+            await asyncio.sleep(0)
+            task.cancel()
+            with self.assertRaises(asyncio.CancelledError):
+                await task
+
+            # After cancellation the socket must be detached (fd == -1)
+            # so that its __del__ won't close a recycled fd.
+            self.assertEqual(sock.fileno(), -1)
+
+        def _recv_or_abort(sock):
+            try:
+                sock.recv_all(1)
+            except ConnectionAbortedError:
+                pass
+
+        with self.tcp_server(_recv_or_abort,
+                             max_clients=1,
+                             backlog=1) as srv:
+            self.loop.run_until_complete(client(srv.addr))
+
+    def test_create_connection_sock_cancel_fd_leak(self):
+        # Regression test for https://github.com/MagicStack/uvloop/issues/645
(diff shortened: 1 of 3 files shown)


junjzhang:
@1st1 Please have a look! thx

6matt:
@fantix could you review? this seems like a critical issue under high concurrency

tarasko:
Hey @junjzhang @6matt, perhaps one option for you would be to switch to [aiofastnet](https://github.com/tarasko/aiofastnet)? Not only it is faster, its networking source code is following python 3.14 asyncio implementation, which doesn't have this bug.

And if there is any other bug, I can fix it and release much faster.

I waited for more than a year to get some of my uvloop PRs merged, and they haven't been released yet.

Just saying

jauntysankey:
I think this is a duplicate of #646

fantix:
> I think this is a duplicate of #646

Good call! Though, #646 fixed it in a way that breaks [asyncio compatibility](https://github.com/python/asyncio/pull/449#issuecomment-255145777). I think this PR is a better approach. Let me also add a theoretical test to reproduce #645, as well as aio-libs/aiohttp#10506.

fantix (approved):
@1st1 wdyt?

jauntysankey:
> > I think this is a duplicate of #646
> 
> Good call! Though, #646 fixed it in a way that breaks [asyncio compatibility](https://github.com/python/asyncio/pull/449#issuecomment-255145777). I think this PR is a better approach.

Thanks for the background. I think the #646 approach was to make passed in socket (PSO) handling more like internally generated socket handling. When libuv generates the socket (such as through an `accept` call), there is no PSO so uvloop creates an internal wrapper. As a result, there are code paths that have to distinguish between a PSO and an internally-wrapped socket. The #646 change made everything just an internally wrapped socket.

fantix:
Yeah, what you said is indeed an ideal resolution, but it is no longer identical to asyncio - where the passed-in `sock` object is still accessible by the caller, no matter how much it is suggested not to be messed with. The referenced asyncio PR was to forbid such wrong usages by re-wrapping the given `sock` object, but that PR was never landed in upstream, that's why uvloop had all those `transport._attach_sockobj()` workarounds for years.

itsdarshankumar:
Hey first of all thanks for the fix @junjzhang 
Can we get an approximate timeline @fantix when this will be released?
Thanks

adamcavendish:
We hit what appears to be the same failure mode in a production disaggregated inference service using Python 3.12, uvloop 0.22.1, aiohttp 3.14.1, and Uvicorn.

Under concurrent aiohttp connection attempts, retries, and cancellation, the API process first returned readiness, then its connections were reset, and the process terminated natively with SIGABRT (`Aborted (core dumped)`). The timing and descriptor-reuse behavior are consistent with the stale Python socket object closing an fd already closed by libuv and later recycled into another transport. The resulting outage produced thousands of downstream 503 responses.

We have switched the affected proxy and Python API serving paths to the standard-library asyncio loop in our release candidate. We plan to keep uvloop disabled until a release containing `dc680eb20c526d8babacaf4751a12f51431ea246` is available and passes our cancellation stress and production-shaped load checks.

Thank you for fixing this and for adding the TCP and Unix socket regression coverage. An approximate release timeline would be very helpful for downstream users deciding when it is safe to re-enable uvloop.

PR #747, opened 2026-04-24
support eager_start and **kwargs in create_task, matching CPython 3.13 to 3.14

### LINKED ISSUE #746: support **kwargs on loop.create_task like in 3.13 and 3.14
3.13.3 — kwargs passthrough introduced (factories required to pass all kwargs on).
3.13.4 — rolled back passing name and context (when None) to avoid breaking backwards compatibility with 3.13.3, while still passing on other arbitrary keyword arguments [Python](https://docs.python.org/3/library/asyncio-eventloop.html). name is no longer passed to task factories, and context isn't passed when it's None.
3.14 — all kwargs are now passed on, and the eager_start parameter works with eager task factories [Python](https://docs.python.org/3/library/asyncio-eventloop.html).

graingert:
…1–3.14

loop.pyx.in is a new Tempita template that emits the correct create_task signature for each Python version at build time:
- 3.14+: `def create_task(self, coro, **kwargs)` forwarding all kwargs to the factory/Task, matching CPython gh-128552
- 3.13:  `def create_task(self, coro, *, name=None, context=None, **kwargs)` passing extra kwargs through to the factory and Task
- 3.11/3.12: `def create_task(self, coro, *, name=None, context=None)` matching CPython's behaviour for those versions
- <3.11:  legacy context.run path

setup.py uses Cython.Tempita to render loop.pyx.in → loop.pyx before calling cythonize, using the process_tempita pattern.

Fixes: https://github.com/MagicStack/uvloop/issues/746

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,8 +94,8 @@ class uvloop_build_ext(build_ext):
 
         for extension in self.distribution.ext_modules:
             for i, sfile in enumerate(extension.sources):
-                if sfile.endswith('.pyx'):
-                    prefix, ext = os.path.splitext(sfile)
+                if sfile.endswith('.pyx') or sfile.endswith('.pyx.in'):
+                    prefix = sfile[:-len('.pyx.in')] if sfile.endswith('.pyx.in') else sfile[:-len('.pyx')]
                     cfile = prefix + '.c'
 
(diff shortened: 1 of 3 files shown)


graingert:
ah upstream CI is failing will close for now

Issue #715, opened 2026-01-14
uvloop debug mode crashes on Python 3.14 during async generator finalization (traceback.walk_stack sees non-frame object)


windsornguyen:
## Overview

When an async generator is finalized under uvloop debug mode, uvloop’s extract_stack() calls traceback.walk_stack() on Python 3.14. The stack walker receives a non-frame object (e.g. _asyncio.TaskStepMethWrapper) and crashes (traceback.walk_stack_generator) or raises AttributeError. It's worth mentioning this does NOT happen with `asyncio` debug.

## Environment

  - uvloop 0.22.1
  - Python 3.14.0
  - macOS (Darwin 24.6.0, arm64)
  - Debug mode enabled (PYTHONASYNCIODEBUG=1 or uvloop.run(debug=True))

## Reproduction

1. Set `PYTHONFAULTHANDLER=1 PYTHONASYNCIODEBUG=1` as environment variables.
2. Run the following script with `python3 uvloop_python314_bug_repro.py`

```python
"""uvloop_python314_bug_repro.py

Minimal repro: uvloop 0.22.1 + Python 3.14 debug async-gen finalization crash.

This creates a nested async-generator chain and breaks on a "[DONE]" sentinel
without closing the inner iterator. Under uvloop debug, async generator
finalization schedules aclose() via call_soon_threadsafe, which calls
extract_stack(). On Python 3.14, traceback.walk_stack sees a non-frame object
(for example, _asyncio.TaskStepMethWrapper) and raises AttributeError. The
unpatched build can segfault in traceback.walk_stack_generator.

Run:
  uv run --active --no-sync python uvloop_python314_bug_repro.py
"""

from __future__ import annotations

import asyncio
import gc
import os
import sys
from typing import AsyncIterator

import uvloop

os.environ["PYTHONASYNCIODEBUG"] = "1"

REPEATS = 1


class FakeResponse:
    async def aiter_raw(self) -> AsyncIterator[bytes]:
        payloads = [
            b"data: one\n\n",
            b"data: [DONE]\n\n",
            b"data: trailing\n\n",
        ]
        for payload in payloads:
            yield payload
            await asyncio.sleep(0)

    async def aclose(self) -> None:
        await asyncio.sleep(0)


class Decoder:
    async def _aiter_chunks(self, source: AsyncIterator[bytes]) -> AsyncIterator[bytes]:
        async for chunk in source:
            yield chunk

[...]

PR #756, opened 2026-07-20
Fix debug-mode segfault when the asyncgen finalizer hook runs during frame teardown

david-runlayer:
## Summary

In debug mode, `Loop._asyncgen_finalizer_hook` can segfault the interpreter. The hook is invoked by the GC from a dealloc — which can happen **while CPython is mid-way through clearing the current interpreter frame** — and the `call_soon_threadsafe` it performs creates a `Handle` whose debug source-traceback capture calls `sys._getframe()`, dereferencing the half-cleared frame.

We hit this in production on CPython 3.13.14 with uvloop 0.22.1: enabling `loop.set_debug(True)` on a Starlette/FastAPI service with streaming request bodies (`request.stream()` is an async generator; under disconnect/cancellation churn they are GC'd constantly) produced worker SIGSEGVs within seconds. Reproduces on 3.13.14 and 3.14.4, x86_64 and aarch64. It appears to be the same underlying weakness as #715 (there, on 3.14, the capture path surfaces a non-frame object to `traceback.walk_stack`).

## Symbolized backtrace (uvloop built `-O0 -g3`, CPython 3.13.14)

```
#3  ?? libpython3.13 (sys._getframe impl; SEGV reading frame->frame_obj)
#4  ?? libpython3.13
#5  __pyx_f_6uvloop_4loop_extract_stack ()                    at cbhandles.pyx (extract_stack)
#6  __pyx_f_6uvloop_4loop_6Handle__set_loop ()                at cbhandles.pyx (Handle._set_loop)
#7  __pyx_f_6uvloop_4loop_new_Handle ()
#8  __pyx_pf_6uvloop_4loop_4Loop_14call_soon_threadsafe ()
#11 __pyx_pf_6uvloop_4loop_4Loop_140_asyncgen_finalizer_hook ()
#14 PyObject_CallOneArg
#16 PyObject_CallFinalizerFromDealloc          <- asyncgen GC'd from a dealloc...
#19 _PyEval_FrameClearAndPop                   <- ...while this frame is being cleared
#20 _PyEval_EvalFrameDefault
```

A production core dump shows the same failure from the release wheel: SIGSEGV (SEGV_MAPERR) inside `sys._getframe` incref'ing `frame->frame_obj` containing a torn value (`0x3000000010`).

## Fix

- `_asyncgen_finalizer_hook` sets a per-thread flag (the hook can run on any thread) so `extract_stack()` skips frame introspection for the handle it creates. The source traceback of a GC point is arbitrary victim code anyway, so nothing meaningful is lost.
- `extract_stack()` additionally tolerates `AttributeError`/`ValueError`/`TypeError` from `StackSummary.extract`/`walk_stack`, returning `None` instead of propagating — this also addresses the 3.14 failure mode reported in #715 (`_asyncio.TaskStepMethWrapper` lacking `f_back`).

## Reproducer

Crashes in <15 s unpatched (verified end-to-end from a clean directory containing only these files: 271 `Fatal Python error` lines in a 60 s window, current PyPI starlette/fastapi/uvicorn); clean with this fix (10 min, ~259k streamed requests, verified on 3.13.14 aarch64 and x86_64).

<details>
<summary>app.py + load.py + docker command</summary>

```python
# app.py — needs: pip install uvicorn starlette fastapi httpx uvloop
from fastapi import FastAPI, Request
from starlette.middleware.base import BaseHTTPMiddleware

app = FastAPI()

class Passthrough(BaseHTTPMiddleware):
    async def dispatch(self, request, call_next):
        return await call_next(request)

app.add_middleware(Passthrough)
app.add_middleware(Passthrough)

@app.post("/upload")
async def upload(request: Request) -> dict:
    total = 0
    async for chunk in request.stream():
        total += len(chunk)
    return {"received": total}

# loop_factory.py
import uvloop
def new_event_loop():
    loop.set_debug(True)          # <- the trigger
    return loop
```

```python
# load.py — 32 concurrent chunked uploads in a loop
import asyncio, httpx

async def body():
    for _ in range(200):
        yield b"x" * 2048

async def worker(client):
    while True:
        try:
            await client.post("http://localhost:8000/upload", content=body())
        except Exception:
            await asyncio.sleep(0.05)

async def main():
    async with httpx.AsyncClient(timeout=30) as c:
        await asyncio.gather(*[worker(c) for _ in range(32)])

asyncio.run(main())
```

```sh
docker run --rm -v $PWD:/r python:3.13.14-slim sh -c '
  pip install -q uvloop==0.22.1 uvicorn starlette fastapi httpx httptools
  cd /r
  PYTHONFAULTHANDLER=1 uvicorn app:app --workers 4 --port 8000 \
    --loop loop_factory:new_event_loop --http httptools &
  sleep 8
  python load.py &
  sleep 60
  # -> "Fatal Python error: Segmentation fault" within ~15s of load starting
'
```

</details>

## Performance

Stock = master merge base (e8efea4), patched = this branch (including the save/restore hook); both built from the same tree with identical flags (aarch64, CPython 3.13.14, best-of-3):

| path | stock | patched | delta |
|---|---|---|---|
| `call_soon` + cancel, non-debug | 0.125 µs | 0.129 µs | noise (≤4 ns, sign flips between sessions) |
| asyncgen create→finalize, non-debug | 1.491 µs | 1.631 µs | +140 ns (three thread-local ops in the hook) |
| `call_soon` + cancel, debug | 3.337 µs | 3.376 µs | ~1% (within noise) |
| asyncgen create→finalize, debug | 11.836 µs | 7.368 µs | −38% (skips the now-pointless capture) |

Non-debug hot paths are untouched by construction (`extract_stack` is only reachable under `loop._debug`); the only non-debug cost is the thread-local save/set/restore per GC'd async generator, on a path whose real cost is dominated by `uv_async_send` plus the subsequently scheduled `create_task`/`aclose()`.

## Behavior change

In debug mode, the handle scheduling `agen.aclose()` from the finalizer hook no longer carries `_source_traceback`. Previously it captured the stack of whichever code happened to be running when GC fired — misleading rather than useful. Task-level debug tracebacks (`create_task` runs later in normal loop context) and `firstiter` capture are unaffected.

Validated with the reproducer plus targeted checks (normal handles still get `_source_traceback` in debug mode; asyncgen finalizer path executes cleanly); relying on CI for the full test matrix.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -720,6 +720,60 @@ class _TestBase:
         self.loop.run_until_complete(foo())
         self.loop.run_until_complete(asyncio.sleep(0.01))
 
+    def test_asyncgen_finalizer_hook_no_stack_capture(self):
+        # The finalizer hook can be invoked from a dealloc while the
+        # interpreter is mid-way through clearing a frame, where debug-mode
+        # source-traceback capture would walk a half-cleared frame and
+        # crash. Handles created while the suppression flag is set must not
+        # capture a traceback; normal handles must keep capturing.
+        if self.implementation != 'uvloop':
+            raise unittest.SkipTest('uvloop only')
+
+        from uvloop.loop import _extract_stack_disabled
+
+        self.loop.set_debug(True)
+        try:
+            handle = self.loop.call_soon(lambda: None)
+            self.assertIsNotNone(handle._source_traceback)
+            handle.cancel()
+
+            _extract_stack_disabled.flag = True
+            try:
+                handle = self.loop.call_soon(lambda: None)
+                self.assertIsNone(handle._source_traceback)
+                handle.cancel()
+            finally:
+                _extract_stack_disabled.flag = False
+        finally:
+            self.loop.set_debug(False)
+
+    def test_asyncgen_finalizer_hook_flag_reentrant(self):
+        # The hook must save/restore the suppression flag, not set/clear it:
+        # allocations inside the hook can trigger a GC that re-enters it on
+        # the same thread, and a clobbering reset would reopen the unsafe
+        # window for the outer invocation.
+        if self.implementation != 'uvloop':
+            raise unittest.SkipTest('uvloop only')
+
+        from uvloop.loop import _extract_stack_disabled
+
+        async def waiter():
+            yield 1
+
+        async def make_agen():
+            agen = waiter()
+            await agen.asend(None)
+            return agen
+
+        agen = self.loop.run_until_complete(make_agen())
+        _extract_stack_disabled.flag = True  # simulate an outer invocation
+        try:
+            self.loop._asyncgen_finalizer_hook(agen)
+            self.assertTrue(_extract_stack_disabled.flag)
+        finally:
+            _extract_stack_disabled.flag = False
+        self.loop.run_until_complete(asyncio.sleep(0.01))
+
     def test_inf_wait_for(self):
         async def foo():
             await asyncio.sleep(0.1)
--- a/uvloop/cbhandles.pyx
+++ b/uvloop/cbhandles.pyx
@@ -416,10 +416,23 @@ cdef new_MethodHandle3(Loop loop, str name, method3_t callback, object context,
     return handle
 
 
+class _ExtractStackDisabled(threading_local):
+    # Per-thread: True while handles are created from the asyncgen finalizer
+    # hook, which a dealloc can invoke while CPython is mid-way through
+    # clearing the current frame; sys._getframe() would then dereference the
+    # half-cleared frame and crash the interpreter.
+    flag = False
+
+
+_extract_stack_disabled = _ExtractStackDisabled()
+
+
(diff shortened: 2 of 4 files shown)


david-runlayer:
CI note: the `test (3.14t, macos-latest)` failure is the pre-existing flake family, not this change — `Test_AIO_Process.test_process_streams_redirect` (stdlib-asyncio variant; this patch isn't in that code path) failed because a debug-mode `Executing <Task ...> took N seconds` slow-callback warning leaked into the redirected stderr on the slow free-threaded runner. The same warning-pollution appears in master's latest Tests run ([29349960216](https://github.com/MagicStack/uvloop/actions/runs/29349960216), 3.14t job, `test_process_delayed_stdio*`), which failed before this PR existed. Re-pushed to retrigger.

zcqian:
I also ran into this issue in `uvloop` and I can confirm your PR fixes the segfaults that I was seeing.

PR #761, opened 2026-08-31
Preserve file descriptor blocking mode in add_reader and add_writer

### LINKED ISSUE #712: inconsistent file descriptor blocking state when running in asyncio vs uvloop
When running under uvloop, registering `stdin`, `stdout`, or `stderr` with `loop.add_reader()` / `loop.add_writer()` implicitly changes their blocking mode from blocking to non-blocking. This behaviour does not occur under the standard asyncio selector event loop.

This is causing some issues when running with rich/promptkit in [fast-agent](https://github.com/evalstate/fast-agent)

Under the following test code

```python
import anyio
import sys
import os
import asyncio

async def main():

    print(f"Using event loop: {type(loop)}")

    stdin = sys.stdin.fileno()
    stdout = sys.stdout.fileno()
    stderr = sys.stderr.fileno()

    print(f"initial:\tstdin fd: {stdin}, stdout fd: {stdout}, stderr fd: {stderr}")
    print(f"blocking:\tstdin fd: {os.get_blocking(stdin)}, stdout fd: {os.get_blocking(stdout)}, stderr fd: {os.get_blocking(stderr)}")

    loop.add_writer(stdin, lambda: None)
    loop.add_writer(stdout, lambda: None)
    loop.add_writer(stderr, lambda: None)

    print(f"writer: \tstdin fd: {stdin}, stdout fd: {stdout}, stderr fd: {stderr}")
    print(f"blocking:\tstdin fd: {os.get_blocking(stdin)}, stdout fd: {os.get_blocking(stdout)}, stderr fd: {os.get_blocking(stderr)}")

    loop.add_reader(stdin, lambda: None)
    loop.add_reader(stdout, lambda: None)
    loop.add_reader(stderr, lambda: None)

    print(f"reader: \tstdin fd: {stdin}, stdout fd: {stdout}, stderr fd: {stderr}")
    print(f"blocking:\tstdin fd: {os.get_blocking(stdin)}, stdout fd: {os.get_blocking(stdout)}, stderr fd: {os.get_blocking(stderr)}")

anyio.run(main, backend="asyncio", backend_options={"use_uvloop": False})
print("-" * 40)
anyio.run(main, backend="asyncio", backend_options={"use_uvloop": True})
```

I see the following output, where the blocking status of the stdio file descriptors are changed from blocking to non blocking when running under uvloop but not asyncio.

```
Using event loop: <class 'asyncio.unix_events._UnixSelectorEventLoop'>
initial:        stdin fd: 0, stdout fd: 1, stderr fd: 2
blocking:       stdin fd: True, stdout fd: True, stderr fd: True
writer:         stdin fd: 0, stdout fd: 1, stderr fd: 2
blocking:       stdin fd: True, stdout fd: True, stderr fd: True
reader:         stdin fd: 0, stdout fd: 1, stderr fd: 2
blocking:       stdin fd: True, stdout fd: True, stderr fd: True
----------------------------------------
Using event loop: <class 'uvloop.Loop'>
initial:        stdin fd: 0, stdout fd: 1, stderr fd: 2
blocking:       stdin fd: True, stdout fd: True, stderr fd: True
writer:         stdin fd: 0, stdout fd: 1, stderr fd: 2
blocking:       stdin fd: False, stdout fd: False, stderr fd: False
reader:         stdin fd: 0, stdout fd: 1, stderr fd: 2
blocking:       stdin fd: False, stdout fd: False, stderr fd: False
```

HyunsikParker:
Fixes #712.

libuv's `uv_poll_init()` switches the supplied file descriptor to nonblocking mode. uvloop left that change in place after `add_reader()` or `add_writer()`, while the asyncio selector loop preserves the original mode.

This records the mode before initializing the poll handle and restores it once initialization finishes. If restoring it fails, the initialized handle is closed before the error is raised.

The regression test covers reader and writer registration and checks the descriptor both while registered and after removal. The same test runs against uvloop and the asyncio reference loop.

Tests:

- `.venv/bin/python -m unittest -v tests.test_sockets` (35 passed, 1 skipped)
- With `test_fs_event_rename` excluded, 528 tests passed with `PYTHONASYNCIODEBUG=1` and again with `PYTHONASYNCIODEBUG=0` (39 skipped in each run)
- The full suite also exposed the existing intermittent macOS `test_fs_event_rename` failure. Isolated reruns produced one pass and one `CHANGE` versus `RENAME` failure.


HyunsikParker:
Two Python 3.10 jobs failed. Neither failure is in the new regression test or the modified poll-handle path:

- Ubuntu: `test_process_delayed_stdio__not_paused__no_stdin` missed the final stdout and connection-lost stages, then reported live process handles.
- macOS: `test_call_later_rounding` measured 50 ms where the test expects at least 69 ms. #755 tracks the early-timer behavior.

The other 14 platform and Python-version jobs passed. I ran both failing tests together 10 times locally, and all 10 runs passed. I do not have permission to rerun upstream Actions jobs. Could a maintainer rerun the two failed jobs?

